Issue #29, opened 2016-05-14
Make :type-at return types of containing expressions if bounds contain no expression


rvion:
I have a module `Test.hs`

```haskell
module Test where

test :: IO ()
test = putStrLn "ok"
```

and I want to use `type-at` to get the type of things I select in my editor, as you can see in the following pictures:

here, I select the letter `S`:
![image](https://cloud.githubusercontent.com/assets/2150990/15269598/107d266e-1a04-11e6-99f0-a157a4fa5f4c.png)

Support this in Intero.

chrisdone:
The `<text>` part is supposed to be for identifiers. Let's include this in #28 as "make better docs". 

It's for when you've moved your code around but you want to get type info of e.g. `putStrLn`, if the location is out of date and the string is now `rLn "blah` then that's no type info. Instead, you provide the alt string so that it can lookup that.

In the case of the user typing it out, we can make it an optional argument.

rvion:
@chrisdone: the <text> part is not the main problem I see.

The main problem I have in this issue is that __`:type-at` doesn't return type of sub-expressions as shown in the image.__ (regardless of text)


the real title of this issue should have been __:type-at doesn't return the type of the closest enclosing expression, but fails when given a sub-expression__

Am I missing something ? 
Should we re-open the issue ?

regarding making "text" optional, I agree.

chrisdone:
Okay, I'll edit the issue and re-open.

rvion:
in other words, my problem is that I can't reproduce pictures above (produced with `ghc-mod`) with `intero` because `intero` doesn't seem to be able to handle sub-expressions.

Having the ability to get instant feedback in an editor when selection change allow powerfull (mouse-free) workflow. 

eg: it gives you the ability to quickly expand the selection to the left while coding to see the type of next arguments you need to provide.

rvion:
@chrisdone I see you edited the original message. 
(I have no problem with that, feel free to edit / remove messages from this thread to clean it up)

I just want to add that the picture you kept is rather easy to implement, but there is more complex cases like this one:

![image](https://cloud.githubusercontent.com/assets/2150990/15269736/16834ab6-1a09-11e6-8535-382f2131650f.png)
(here, I select `" "`)

chrisdone:
Sure, I understood the meaning of the issue now. Intero is stricter because I use structured-haskell-mode which can select the exact bounds of an expression on a key press. But it's pretty straight-forward to add this for "free selection". 

[...]

PR #30, opened 2016-05-14
[RDY] Fix 2 more bugs related to #29

rvion:
@chrisdone 

While investigating the 2 more "incorrect" results given by :type-at (found when reproducing examples pictures in #29), I grew the intuition that `inside` was not needed at all in src/GhciFind.hs.

- :type-at for the “f a” part of “f a b c” was returning the type of f
- :type-at for the “a b” part of “f a b c” was returning the the type of a
- logic was duplicated 

Indeed, `contains` seems to be the correct logic in all cases, because we always want to find the smallest enclosing type (that was also the behaviour on the web based fpcomplete haskell center ide)

Previsouly, `inside` was always matching the special case where child and parent were the same.

I'll add adding more tests to check that it behave as ghc-mod, but I think the logic is better now.
(the test suite is passing on my laptop).

### DIFF
--- a/src/GhciFind.hs
+++ b/src/GhciFind.hs
@@ -266,8 +266,8 @@ resolveSpanInfo :: [SpanInfo] -> Int -> Int -> Int -> Int -> Maybe SpanInfo
 resolveSpanInfo spanList parentSL parentSC parentEL parentEC =
   find contains spanList
   where contains (SpanInfo ancestorSL ancestorSC ancestorEL ancestorEC _ _) =
-          ((ancestorSL == parentSL && parentSC >= ancestorSC) || (ancestorSL > parentSL)) &&
-          ((ancestorEL == parentEL && parentEC <= ancestorEC) || (ancestorEL < parentEL))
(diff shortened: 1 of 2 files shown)


rvion:
edit: ⚠️ the implicit invariants resolveSpanInfo is relying on makes me a bit inconforable. I think there is some problems with the current code because spans order in modinfoSpans is not always reliable (I am seeing problems with multi-line selections and do blocks)

rvion:
travis error is due to 
> The command "travis_retry curl -L https://www.stackage.org/stack/linux-x86_64 | tar xz --wildcards --strip-components=1 -C ~/.local/bin '*/stack'" failed and exited with 2 during .

rvion:
@chrisdone I'm ready for review :)

As a side note, I plan to make a small web based haskell ide on top of intero soon.

chrisdone:
Cool. Taking a look now :)

chrisdone:
Test cases look good. I squashed the commits and linked up the test cases to issues. Pushed as 4af84bf. Cheers!

PR #32, opened 2016-05-16
Don't use -dynamic on Windows

lukexi:
`-dynamic` doesn't work on Windows. Intero works quite nicely with this in though!


chrisdone:
😄 👍 Nice one!

chrisdone:
What version of GHC are you using, by the way?

lukexi:
I'm on 7.10.3

PR #33, opened 2016-05-19
add configuration completion

obadz:
`:completion` command can be used to set `completionMode` for a given macro

```
syntax: :completion <macro> none|expr|file|module|identifier
```
Example:

```haskell
:def which-package \ m -> return $ ":! ghc-pkg --simple-output find-module " ++ m
:completion which-package module
```


obadz:
Another example:

```haskell
λ> :def cat \s -> return $ ":!cat " ++ s
λ> :completion cat file
λ> :cat .gitig[Tab]
```

chrisdone:
Nice!

Issue #37, opened 2016-05-22
Type info not collected for `:add`-ed modules


mgsloan:
Repro:

1) Run `stack ghci --with-ghc intero` in the stack repo (perhaps with `--no-build` if it's already been built)
2) Try `:all-types` and get no results
3) Do a `:load Main`
4) Now `:all-types` yields all types

The difference is that the ghci-script generated by stack uses `:add`.

mgsloan:
Opened a PR which seems to address this.

chrisdone:
Think we can close this now.

PR #38, opened 2016-05-22
Collect typeinfo no matter the reason for load #37

mgsloan:


chrisdone:
Cool, will merge when I'm back at a computer.

PR #51, opened 2016-06-03
Strip properties from (buffer-file-name).

mboes:
Some modes decorate the buffer file name with faces, e.g. flx-ido.
Arguably this is a bug in said modes, but at any rate this simple fix
protects against any potential issues.

In particular, if we don't strip properties, then the `:type-at`
string generated by intero includes garbage and a syntax error ensues.

### DIFF
--- a/elisp/intero.el
+++ b/elisp/intero.el
@@ -67,7 +67,7 @@
 (define-minor-mode intero-mode "Minor mode for Intero"
   :lighter " Intero"
   :keymap intero-mode-map
-  (when (buffer-file-name)
+  (when (intero-buffer-file-name)
     (if intero-mode
         (progn (flycheck-select-checker 'intero)
                (flycheck-mode)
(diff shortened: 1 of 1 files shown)


chrisdone:
I wasn't aware of that! Thanks!

polux:
This broke the intero layer in spacemacs for me. I get `(wrong-type-argument stringp nil)` when typing C-c C-t and the stack trace is:

  substring-no-properties(nil)
  intero-buffer-file-name()
  intero-mode()
  run-hooks(change-major-mode-after-body-hook prog-mode-hook haskell-mode-hook)
  apply(run-hooks (change-major-mode-after-body-hook prog-mode-hook haskell-mode-hook))
  run-mode-hooks(haskell-mode-hook)
  haskell-mode()
  intero-type-at(nil)
  call-interactively(intero-type-at nil nil)
  command-execute(intero-type-at)

chrisdone:
Oh, you're right. That'd do it. I'll fix it.

chrisdone:
@polux Pushed. Does that fix it?

polux:
Yes it does! Thanks!

PR #52, opened 2016-06-04
Nix support [WIP]

acowley:
I think this works for me, but it needs some more banging on it. I've also not yet verified that I haven't changed behavior when using `stack`. Hopefully another person or two can give it a shot.

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -1,0 +1,29 @@
+{ mkDerivation, array, base, bytestring, containers, directory
+, filepath, ghc, ghc-paths, haskeline, hspec, process, regex-compat
+, stdenv, syb, temporary, time, transformers, unix
+}:
+mkDerivation {
+  pname = "intero";
+  version = "0.1.13";
+  src = builtins.filterSource
+          (path: type: let baseName = baseNameOf path; in with stdenv.lib;
+           !(hasSuffix "~" baseName || hasPrefix ".#" baseName || (type == "directory" && baseName == "elisp")))
+          ./.;
+  isLibrary = false;
+  isExecutable = true;
+  prePatch = ''
+    sed -i 's|elisp/*.el||' ./intero.cabal
+    sed -i 's|data-files:||' ./intero.cabal
+  '';
+  executableHaskellDepends = [
+    array base bytestring containers directory filepath ghc ghc-paths
+    haskeline process syb time transformers unix
+  ];
+  testHaskellDepends = [
+    base directory hspec process regex-compat temporary transformers
+  ];
+  doCheck = false;
+  homepage = "https://github.com/commercialhaskell/intero";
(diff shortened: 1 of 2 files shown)


chrisdone:
Thanks for the PR, Anthony!

The PR seems a bit messed up; it duplicates code (now there're two `define-minor-mode`) somehow. Must be a merge artifact?

I'm probably not going to merge Nix support into master at this time in the project's life cycle, for the same reasons I didn't write the patch necessary to work with cabal "sandboxes":

* Stack has a notion of projects, which contain packages, with targets.
* Stack allows automatic installation of things.
* Stack is going to provide us with information (like which targets are available, flags, etc.) either as an executable or a linked library.
* Things like `stack hoogle` in the future are going to be key pieces of functionality.

In other words, we assume Stack, it does things for us and that helps make things work out of the box.

*One more way* of doing all this stuff is *exactly* the kind of thing that makes `haskell-mode` proper complicated and difficult to debug (and fragile to maintain; everything merged into `haskell-mode` lately seems to break something I'm using--there's a motivation behind not depending on anything in `haskell-mode` directly from `intero.el`). 😭  Trying to please everybody at the risk of breaking a smooth user experience on the intended way of doing things leads to configuration, and every time a user has to enter a configuration option to get basic functionality, we've failed as a user experience. ☝️ 

However, I wouldn't mind if you just fork intero for now and work on your own branch that irons out the kinks in the nix support. I'm not sure whether it could be a separate mode like `intero-nix`, but if anything gets in I'd like it to be first-class support, not redefining variables or whatnot like often done in `haskell-mode`. Is it possible to start your Emacs inside `nix-shell`, as a compromise?

mboes:
Note: just yesterday (before this PR was submitted) I sneaked in the following two lines to the stack.yaml:

```yaml
nix:
  packages: [ ncurses ]
```

This fixes intero compilation on NixOS for me. And shows one way to get a `nix-shell` but via Stack, therefore without requiring any special support from intero: you just need to add a `nix:` section to your project's stack.yaml. Could this solve your use case?

chrisdone:
@mboes That would be an ideal scenario.

acowley:
Yeah, there was a nasty merge at the end there that must have picked up some cruft. I can use my fork until something else comes along, but the difficulty I already had tracking changes doesn't bode well for maintenance. Thanks for taking a look.

sleexyz:
one hack I did to get intero to install automatically via MELPA is load emacs from a nix-shell with ncurses:

```
$ nix-shell -p ncurses
...
[nix-shell:~]$  emacs Main.hs
```

puffnfresh:
Intero doesn't use the `--nix` Stack flag, so I forced it via `~/.stack/config.yaml`:

```yaml
nix:
  enable: true
```

PR #54, opened 2016-06-04
Don't print "Wrote <file>" message on flycheck

abaw:
In defun intero-check, calling write-region will print a message "Wrote
<file>". In my environment, flycheck always happens after doing "C-c
C-t". And the "Wrote <file>" makes the type information in my minibuffer
gone in one second or so. This is really annoying.


abaw:
Just checked intero and found it's useful. But this minor issue is annoying.

chrisdone:
You have no idea how long I looked for a way to disable that message! I bow down to your Emacs help screen prowess!

abaw:
It's good to know this helps.

Issue #57, opened 2016-06-04
Windows file names: \ vs /


lukexi:
Giving Intero a whirl on Windows - many things seem to work (`C-c C-t`, completion) but I haven't managed to get any flycheck output yet. Best guess so far is that it's a line ending issue, as the strings coming back from intero-async-call are chock full of `^M`s when I dump them. 

I don't use emacs regularly so I'll be a bit slow to debug it; will try though! Also happy to test any ideas.

chrisdone:
Interesting. How about printing out some of the string values like `string` in here? https://github.com/commercialhaskell/intero/blob/master/elisp/intero.el#L268..L274

Perhaps the ` "OK, modules loaded: \\(.*\\)\\.$"` regex is too strict. Try removing the `$`?

lukexi:
I got a little further via stripping with
`(let ((string (replace-regexp-in-string "\r" "" (buffer-substring (point-min) (1- (point))))))`
at 
https://github.com/commercialhaskell/intero/blob/master/elisp/intero.el#L913

Which gets me past "OK, modules loaded" and I'm able to use C-c C-i now (before I'd get "not in scope" errors).

I'm inspecting the results of `intero-parse-errors-warnings` now and they seem to be constructing correctly... but FlyC reports nothing when asking it to show all errors (these are all warnings, but I've tried inducing errors as well).
```
([cl-struct-flycheck-error nil intero "C:\\msys64\\home\\lukex\\Projects\\rumpus\\src\\Rumpus\\Systems\\Hands.hs" 98 1 "
    Top-level binding with no type signature:
      isBeingHeldByHand :: forall (m :: * -> *).
                           (MonadState ECS m, MonadReader EntityID m) =>
                           WhichHand -> m Bool" warning nil] [cl-struct-flycheck-error nil intero "C:\\msys64\\home\\lukex\\Projects\\rumpus\\src\\Rumpus\\Systems\\Hands.hs" 8 1 "
    The qualified import of ‘Data.List’ is redundant
      except perhaps to import instances from ‘Data.List’
    To import instances alone, use: import Data.List()" warning nil] [cl-struct-flycheck-error nil intero "C:\\msys64\\home\\lukex\\Projects\\rumpus\\src\\Rumpus\\Systems\\Creator.hs" 313 1 "
    Top-level binding with no type signature:
      spawnChildInstance :: forall (m :: * -> *).
                            (MonadIO m, MonadState ECS m, MonadReader EntityID m) =>
[...]

PR #61, opened 2016-06-04
Path and newline normalization for Windows support

lukexi:
Addresses #57


elisp/intero.el
```diff
@@ -624,7 +624,19 @@ May return a qualified name."
   "Return buffer-file-name stripped of any text properties."
   (let ((name (buffer-file-name buffer)))
     (when name
-      (substring-no-properties name))))
+      (canonicalize-path (substring-no-properties name)))))
+
+(defun canonicalize-path (path)
```
chrisdone: It's not usual to expose top-level functions without a module prefix (like `-intero`), but don't worry I'll add that before merging!

chrisdone:
Merged, with namespace prefixes! Thanks for the work!

PR #62, opened 2016-06-04
Implement basic eldoc-mode backend

iainnicol:
The use case for this is similar to that of haskell-mode's
haskell-doc-mode.  However, this has the advantage of using intero to
retrieve the type information.


chrisdone:
I was thinking of doing the same thing! Will merge when I get home.

chrisdone:
Thanks!

iainnicol:
:)

PR #68, opened 2016-06-05
Added default key bindings in readme.

yamafaktory:
I thought it should be great to display the default provided key bindings for Emacs (especially intero-repl-load which is missing in http://commercialhaskell.github.io/intero/).

Thanks for your awesome work!

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,10 +9,10 @@ Please see
 
 ### Default key bindings
 
-. intero-goto-definition `M-.`
-. intero-info `C-c TAB`
-. intero-repl-load `C-c C-l`
-. intero-type-at `C-c C-t`
+- intero-goto-definition `M-.`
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -7,6 +7,13 @@ Complete interactive development program for Haskell
 Please see
 [the homepage for Intero for Emacs](http://commercialhaskell.github.io/intero).
 
+### Default key bindings
+
+- intero-goto-definition `M-.`
+- intero-info `C-c TAB`
```
chrisdone: I think I'll tweak this to `C-c C-i` with `C-c TAB` in brackets. The `-i` is mnemonic for "info". I think TAB won't mean anything for remembering.

chrisdone:
Thanks!

PR #70, opened 2016-06-05
Save current buffer before loading REPL

fosskers:
People used to the `interactive-haskell-mode`  REPL loading behaviour (`C-c C-l`) expect the current buffer to be saved before loading.

To demonstrate the current behaviour, define:

```haskell
foo = putStrLn "hey"
```
Save, `C-c C-l` and test the value of foo. The repl will print `hey`.

Now change `foo` to:

```haskell
foo = putStrLn "hi"
```
and without saving, load the repl again with `C-c C-l`. `foo` will still print `hey`.

### DIFF
--- a/elisp/intero.el
+++ b/elisp/intero.el
@@ -469,6 +469,7 @@ warnings, adding CHECKER and BUFFER to each one."
 (defun intero-repl-load ()
   "Load the current file in the REPL."
   (interactive)
+  (save-buffer)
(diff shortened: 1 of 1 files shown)


chrisdone:
Very reasonable! An omission on my part. Thanks!

fosskers:
Fix confirmed with today's `intero` out of melpa.

PR #78, opened 2016-06-06
Push find-tag-marker-ring so alt-* goes back from goto-definition

Peaker:


Blaisorblade:
Oh thanks, this should fix half of #81 (ahem, haven't tested). What about `M-,`? I guess it doesn't make sense with `intero` — should it be overriden to give an error?

k-bx:
@Blaisorblade oh, please don't override `M-,` for an error, I'm still using it with regular TAGS which still work superior to intero sometimes (and can understand different languages or other things interesting for you).

chrisdone:
I'll merge this in a minute. @purcell commented when he merged it into MELPA that I should add this, but he added some extra backwards compatibility in the code. I'll incorporate both things!

chrisdone:
Reference here FYI: https://github.com/melpa/melpa/pull/3959#issuecomment-223711290

chrisdone:
Merged and applied Purcell's suggestion too.

purcell:
Yay, thanks for that!

Issue #46, opened 2016-05-29
Support GHC 8


chrisdone:

chrisdone:
Done.

obadz:
I notice that this merge includes a revert of 7309426641dec9b5a015d666e4a2fe1ab119a7e0 ­-- was that intentional?

PR #93, opened 2016-06-07
reapply 7309426 (configurable completion) which was removed in 9bef3b3

obadz:
removed in #46 by accident?


chrisdone:
Oh, woops. I think I reset too far while doing that.

obadz:
Thanks, do you think we could get a version bump on hackage?

chrisdone:
Pushed 0.1.14.

obadz:
Thanks a lot!

PR #100, opened 2016-06-08
Make tests pass in non-UTF-8 locale (i.e. nix build)

rvl:
Test failed when running `stack test --nix`.


rvl:
Cancelling PR because the change fails on osx.

chrisdone:
Why is nix always so problematic? It seems disproportionate.

rvl:
Seems like a question for `M-x doctor` ... or reddit.

Issue #56, opened 2016-06-04
Support completion for stack targets


chrisdone:
I'd like an interface that pops up something like this:

```
[x] weigh:lib  [ ] weigh:test:weigh-test
[ ] store:lib  [ ] store:test:store-test [ ] store:bench:store-bench
```

And you can use your cursor or mouse to click the targets you want.

For this we need a list of packages from Stack, and a list of available stanzas from within each one, according to what is configured with the current flags, etc. Stack is completely capable of passing this information to us.

magthe:
Maybe this could be done via `helm` or something similar?

chrisdone:
@magthe I'm not up on helm. If you're familiar with it can you answer a couple quick questions:

1) does it have a small foot-print (the kind users wouldn't mind paying; like flycheck and company)?
2) is it trivially programmable, so we could use it for this use-case (and maybe others)?
3) does it take over your whole emacs or is it like ido where it can be used on a per-need basis?

magthe:
I'm not intimately familiar with it either, I only use it via *spacemacs*.

1. I'm not sure about the footprint but `helm-core` and `helm` occupy 608k and 1.6M respectively on my system, so it's a bit larger than `flycheck` and `company`.
2. I believe it's trivially programmable, at least if one looks at the number of extensions to it for various things.
3. It pretty much does take over Emacs.

Given 2 and 3 I guess I should instead request that you don't do it in such a way that it's difficult to integrate with `helm`.

akermu:
As an alternative there's also the stock emacs interface for such things: `completing-read`. Helm and other completion systems, like ivy or ido, hook into it, so users of these could use it as well.

chrisdone:
The `completing-read` function just lets you choose one option at once. We want to choose *n* at once. I.e. we want to enable many stack targets in one session (library and test suite, for example).

magthe:
A possibly completely naive and silly thing to say, but how about `completing-read-multiple` from `crm.el`?

chrisdone:
@magthe That's not bad but I think part of my motivation for this completion is discoverability; if I don't know how to specify stack targets or what mine are called, it doesn't matter; the UI shows the available targets immediately.

chrisdone:
I've pushed an example in 62246d9. It's used like this:

``` lisp
(intero-multiswitch
 "Pick your targets: "
 '((:key foo :title "foo")
   (:key foo-exe-bar :title "foo:exe:bar")
[...]

PR #131, opened 2016-06-28
Add command to start the repl and have stack load all modules in the target

wuzzeb:
The new command `intero-repl-load-all-modules` starts a new repl without the --no-load GHCI flag, so that `stack ghci` loads all the modules for the given target.  It only works if you use the command the very first time you start the repl, but you can quit GHCI and kill the buffer if you want to restart.

### DIFF
--- a/elisp/intero.el
+++ b/elisp/intero.el
@@ -109,6 +109,13 @@
              (if intero-global-mode
                  "enabled" "disabled"))))
 
+(defvar intero-repl-no-load t
+  "When starting the repl, pass --no-load to skip loading the files
+  from the selected target")
+
+(defvar intero-repl-no-build t
+  "When starting the repl, pass --no-build to skip building the target")
+
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 ;; Buffer-local variables/state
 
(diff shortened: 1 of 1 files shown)


chrisdone:
I'm reluctant to add a new command, as this feature could be added to the existing command. But I like the idea to optionally load everything.

Proposal:

With the `C-u` prefix argument before running `intero-repl` or `intero-repl-load`, it brings up a little options menu, similar to magit. This would look a bit [like this](https://github.com/commercialhaskell/intero/issues/56#issue-158493916), so we could re-use the same interface.

    l - [ ] Load all modules   b - [ ] Build project first
    
    d - [X] Load .ghci files   o  - [X] Use -fobject-code

I.e. rather than encoding all options in various commands, pop it in a menu.

chrisdone:
I've now implemented the support for such a UI, if you're interested! https://github.com/commercialhaskell/intero/blob/master/elisp/intero.el#L1372

It's used by `intero-targets`, for example. We could present the above options using this feature.

wuzzeb:
Ok, I implemented a small menu which uses the new intero-multiswitch.  At the moment it just has no-load and no-build, but the other options can easily be supported.  I made the options global which makes the most sense to me, do you think they should be buffer-local instead?

chrisdone:
I'm not sure about the buffer-local vs global. I guess global works fine.

chrisdone:
Merged! I made the changes I mentioned above. So:

1. The two options are customizations that you can configure globally for all future projects.
2. If you run `intero-repl` or `intero-repl-load` with `C-u` (prefix argument), it opens the options prompt. The options that you choose in that prompt are specific to your current project (backend buffer) only. If you restart your REPL and prompt options, it shows the previously selected options for that project.
3. Once you restart Emacs, it reverts back to your global configuration.

PR #151, opened 2016-07-10
Don't misuse debug-on-error to determine verbosity

purcell:
I like to set `debug-on-error` so that things fail noisily -- this doesn't mean I want debug output from everything. This commit introduces an `intero-debug` var to specifically control whether `intero` is verbose.

### DIFF
--- a/elisp/intero.el
+++ b/elisp/intero.el
@@ -83,6 +83,11 @@ target."
   :type 'boolean)
 (make-variable-buffer-local 'intero-repl-no-build)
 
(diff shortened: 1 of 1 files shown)


chrisdone:
We should add a toggle function otherwise it's more verbose to tell people how to get debug info.

purcell:
Yeah, maybe, but that's why I made it a `defcustom` -- it's right there if people browse for it.

chrisdone:
Added `M-x intero-toggle-debug`.

PR #169, opened 2016-07-18
Offer to apply GHC's suggested typo fixes

acowley:


acowley:
A rather significant issue with this is the coupling between the elisp code and the version of GHC. One idea is to have a table of regular expressions indexed by the GHC version. I've been testing this with GHC-8.0.1, but the message parsing will surely break with other versions.

chrisdone:
Oo I look forward to trying this out when I get home tomorrow!

chrisdone:
Not back at desk yet but it looks pretty good to me. I'll merge when I get back. One thing to check is that these rewrites are applied BEFORE adding or removing lines such as in imports and extensions. I can't tell from the mobile app but the loop should come before the other loops obviously. :+1:

chrisdone:
> A rather significant issue with this is the coupling between the elisp code and the version of GHC. One idea is to have a table of regular expressions indexed by the GHC version. I've been testing this with GHC-8.0.1, but the message parsing will surely break with other versions.

Agreed. I'll make such a variable (GHC version) available.

chrisdone:
@acowley I rebased and merged, but changed the algo. to pick up suggestions. Would you mind testing on GHC 8? Hopefully this regex will work on a few at once, and support multiple replacement suggestions.

As a todo item I'll need to create a test suite which will also run Emacs on all the GHC versions, not just intero the executable.

chrisdone:
Thanks for the contribution! 

I'd like to add another one for `Top-level binding with no type signature: main :: IO ()` when I get round to it. 💃

acowley:
The new regex isn't actually even close to working for GHC8.0.1. I can probably merge my working one with the one you added to keep it as a single regex, or we can start down the path of having GHC version specific regexs. Any preferences?

chrisdone:
If it can be simply handled with one regex I'd prefer just one. But the GHC-specific ones are an option later.

PR #172, opened 2016-07-19
Bump auto-installed intero version

Tritlo:
Bump the auto-installed intero version in the elisp file to be the latest version, 0.1.16.

### DIFF
--- a/elisp/intero.el
+++ b/elisp/intero.el
@@ -63,7 +63,7 @@
   :group 'haskell)
 
 (defcustom intero-package-version
-  "0.1.15"
+  "0.1.16"
   "Package version to auto-install."
   :group 'intero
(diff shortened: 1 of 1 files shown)


chrisdone:
What do you need this for?

chrisdone:
Back at my desk. The Emacs mode doesn't use any features in `0.1.16` - why do you need this update? Doing so will force anyone who updates their elisp to reinstall intero.

Tritlo:
Ah, I see. My mistake. I just thought it was weird that the version fetched by Emacs was not in sync with the latest version.

chrisdone:
No worries. I might pop a comment in there explaining this as you're not the first to notice it, good eye.

PR #173, opened 2016-07-19
Fix GHC8 compatibility of "Not in scope" regex

acowley:


acowley:
Closing in favor of #174

PR #174, opened 2016-07-19
Default typo fix

acowley:
This fixes the "not in scope" regex to work with GHC8 and sets a typo fix replacement widget to be enabled by default if it is the only suggested fix.

### DIFF
--- a/elisp/intero.el
+++ b/elisp/intero.el
@@ -1807,7 +1807,7 @@ suggestions are available."
                                       "’ with ‘"
                                       (plist-get suggestion :replacement)
                                       "’")
-                       :default nil))))
+                       :default (null (cdr intero-suggestions))))))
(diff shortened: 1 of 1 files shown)


chrisdone:
>  to be enabled by default if it is the only suggested fix.
Nice! I was hoping to do that but didn't have the time to squeeze it in.

elisp/intero.el
```diff
@@ -1807,7 +1807,7 @@ suggestions are available."
                                       "’ with ‘"
                                       (plist-get suggestion :replacement)
                                       "’")
-                       :default nil))))
+                       :default (null (cdr intero-suggestions))))))
```
chrisdone: Technically you can get more than one `Not in scope` error at once. But I'll follow that up with a patch later.

PR #184, opened 2016-07-22
Apply GHC's suggestions on redundant constraints

acowley:
This one is a bit involved on the parsing side.


chrisdone:
Definitely going to need a test suite but I think this is valuable and
worth investing in. Will check out when I'm home!

PR #185, opened 2016-07-23
add suggestions for turning off name shadowing and default warnings.

tonyday567:
I imagine there are a lot of suggestions to come on line, and things could get long and messy.

Happy to help with a refactor if anyone has a sense of how to handle these cases.

### DIFF
--- a/elisp/intero.el
+++ b/elisp/intero.el
@@ -1715,6 +1715,31 @@ suggestions are available."
               (setq start (min (length text) (1+ (match-end 0))))))
           ;; Messages of this format:
           ;;
+          ;; Defaulting the following constraint(s) to type ‘Integer’
+          ;;   (Num a0) arising from the literal ‘1’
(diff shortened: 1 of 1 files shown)


chrisdone:
Looks good in the diff! I'll back at  keyboard tonight to try it out and
will merge. I agree with your choice of not setting these on as default.

A future additional suggestion could be to use intero's knowledge of
variable uses and offer a rename to e.g. x' or whatnot.

Regarding the blow up of cases let's play it by ear and then refactor once
we're pretty sure most of the cases have been exhausted!

acowley:
This looks good to me, too, but I'd *really* like that ability to rename a shadowing identifier!

chrisdone:
Agreed. I'd be a prime user of such a rename.

Sorry I didn't get to merging yesterday, I got stranded in Verona due to
train track accident downstream.

PR #192, opened 2016-07-26
Added support for compiling under a different working directory than GHCi.

AndrewRademacher:
This is done to address issues with loading multiple packages in intero/GHCi at the same time. Stack will use the new `cd-ghc` command to control the root directory under which packages compile. This allows for TemplateHaskell to behave in intero as it would during a normal GHC build.


chrisdone:
Oh, fun. Let me check it out.

src/InteractiveUI.hs
```diff
@@ -535,29 +536,32 @@ interactiveUI config srcs maybe_exprs = do
...
-                   last_command   = Nothing,
-                   cmdqueue       = [],
-                   remembered_ctx = [],
-                   transient_ctx  = [],
-                   ghc_e          = isJust maybe_exprs,
-                   short_help     = shortHelpText config,
-                   long_help      = fullHelpText config,
-                   mod_infos      = M.empty,
-                   rdrNamesInScope = []
+        GHCiState{ progname            = default_progname,
```
chrisdone: Next time I'd probably put reformatting in another commit. It makes it harder to tell what changed in the github view.

chrisdone:
We could probs add a test in the test suite for this.

src/InteractiveUI.hs
AndrewRademacher: Ahh, the formatting is in a separate commit. But if GitHub is going to merge them all together this way, I'll start doing it in a separate PR. :/

chrisdone: D'oh. No worries, I'll peruse each commit next time.

Issue #189, opened 2016-07-25
Add hyperlinks to repl


angerman:
Can we parse `^<path>:<line>:<char>$` lines in the repl, and turn them into hyperlinks that take you to that relevant buffer and line, or open that file in a new buffer and take you to the line?

chrisdone:
That's possible, sure. Do you know Elisp to take a crack at it?

angerman:
To be quite honest, I do know lisp a bit, but fail on Elisp all the time, because I can't find out how to lookup which functions are available to me.

I'm usually already stuck at `(? ...)`. Do you have any pointers to a elisp function reference?

acowley:
Just in case it helps, I can't recommend `helm` enough. I'm not very good at elisp, but using `helm` to discover things via `M-x` or `M-:` easily doubled whatever ability I brought to the table.

chrisdone:
Check this out https://github.com/chrisdone/elisp-guide

angerman:
Alright, no excuses for me anymore. Thanks @acowley and @chrisdone! I'll try to give this a shot.

angerman:
Ok. I got some basic detection and highlighting working; not really sure how to approach the open file part yet.

```elisp
(defun jump-to-file ()
  (interactive)
  (message "%S:%S" (get-text-property (point) 'file) (get-text-property (point) 'line)))

(defvar make-mouse-map
  (let ((map (make-sparse-keymap)))
    (define-key map [mouse-1]  'jump-to-file)
    (define-key map [return]   'jump-to-file)
    (define-key map [C-return] 'jump-to-file)
    map)
  "Keymap for clicking on links in intero.")

(defun highlight-file:line-references (from to len)
  (save-excursion (let ((text (let ((start (progn (goto-char from) (beginning-of-line) (point)))
                                    (end   (progn (goto-char to) (end-of-line) (point))))
                                (buffer-substring-no-properties start end)))
                        (start 0)
                        (file-regex "^\\([[:alpha:]/]+\\):\\([0-9]+\\):\\([0-9]+\\)$"))
                    (while (string-match file-regex text start)
                      (let ((file (match-string 1 text))
                            (line (match-string 2 text))
                            (link-start (1+ (match-beginning 1)))
                            (link-end   (1+ (match-end 3))))
                        (add-text-properties link-start link-end (list
                                                                  'file file
                                                                  'line line
                                                                  'keymap make-mouse-map
[...]

Issue #194, opened 2016-07-27
Inside Emacs Intero REPL buffer quickBatch spams REPL with ^H characters


AlainODea:
If I load **QuickBatchReplSpamExample.hs** into the Intero REPL within Emacs and run `main`... (continued after code listing)

**QuickBatchReplSpamExample.hs**:

```haskell
module QuickBatchReplSpamExample where

import Control.Applicative
import Test.QuickCheck hiding (Success)
import Test.QuickCheck.Checkers
import Test.QuickCheck.Classes

newtype Identity a = Identity a deriving (Show, Eq)

instance Functor Identity where
  fmap f (Identity x) = Identity $ f x

instance Applicative Identity where
  pure = Identity
  (<*>) (Identity f) = fmap f

instance Arbitrary a => Arbitrary (Identity a) where
  arbitrary = Identity <$> arbitrary

instance Eq a => EqProp (Identity a) where (=-=) = eq

identityTrigger :: Identity (String, String, Int)
identityTrigger = undefined

main :: IO ()
main = do
  quickBatch $ applicative identityTrigger
```

I get output like this:
```plain
identity:     (0 tests)^H^H^H^H^H^H^H^H^H         ^H^H^H^H^H^H^H^H^H(1 test)^H^H^H^H^H^H^H^H         ^H^H^H^H^H^H^H^H(2 tests)^H^H^H^H^H^H^H^H^H         ^H^H^H^H^H^H^H^H^H(3 tests)^H^H^H^H^H^H^H^H^H         ^H^H^H^H^H^H^H^H^H(4 tests)^H^H^H^H^H^H^H^H^H         ^H^H^H^H^H^H^H^H^H(5 tests)^H^H^H^H^H^H^H^H^H         ^H^H^H^H^H^H^H^H^H(6 tests)^H^H^H^H^H^H^H^H^H         ^H^H^H^H^H^H^H^H^H(7 tests)^H^H^H^H^H^H^H^H^H         ^H^H^H^H^H^H^H^H^H(8 tests)^H^H^H^H^H^H^H^H^H
```

Output continues like that for the 500 tests. It also runs a lot slower than it does in Terminal or in eshell/shell buffers (seems to be lagged by Emacs rendering).

The **^H**s are rendered as if there is an encoding glitch. The caret moves from ^ to ^ and copying sections of the output from Emacs does not include these characters, but includes what appear to be an equivalent number of zero-width spaces.

[...]

PR #198, opened 2016-07-27
Partially working jump-to-error

angerman:
This is basically as far as I came. Any input would be much appreciated.
- style
- naming
- ...

The `comint-output-filter-functions` hook is *not* called when a buffer is loaded through `C-c C-l`, which I could not figure out why. It also keep reprocessing the whole buffer, which is why it actually works when hitting `ret` on the prompt after a broken file was loaded...

We might also want to consider setting (and making it configurable)
- `comint-buffer-maximum-size`
- `comint-input-ring-size`

To limit the buffer's size to a sensible amount of history?

### DIFF
--- a/elisp/intero.el
+++ b/elisp/intero.el
@@ -780,9 +780,7 @@ If PROMPT-OPTIONS is non-nil, prompt with an options list."
 (defun intero-linkify-file:line (begin end)
   "Linkify all occurances of <file>:<line>:<char>: betwen begin and end"
   (let ((end-marker (copy-marker end))
-        (file:line:char-regexp (concat "[\r\n]\\([A-Z]?:?[^ \r\n:][^:\n\r]+\\):\\([0-9()-:]+\\):"
-                                       ;; "[ \n\r]+\\([[:unibyte:][:nonascii:]]+?\\)\n[^ ]"
-                                       )))
+        (file:line:char-regexp "[\r\n]\\([A-Z]?:?[^ \r\n:][^:\n\r]+\\):\\([0-9()-:]+\\):"))
     (save-excursion
       (goto-char begin)
       ;; Delete unrecognized escape sequences.
@@ -798,7 +796,8 @@ If PROMPT-OPTIONS is non-nil, prompt with an options list."
                                      'help-echo "mouse-2: visit this file")))))))
 
 (defun intero-linkify-process-output (ignored)
-  ""
+  "comint-output-filter-function to turn <file>:<line>:<char>: into
+links that can be clicked on."
   (let ((start-marker (if (and (markerp comint-last-output-start)
                                (eq (marker-buffer comint-last-output-start)
                                    (current-buffer))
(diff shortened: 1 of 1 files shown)


angerman:
This would potentially fix #189

angerman:
Alright, this seems to be working for me now.

I'd appreciate some input on:
- style
- naming
- how we want to deal with the re? Using the full one used elsewhere in intero does not match, as the incremental output is missing a final new-line, unless I misinterpreted something.
- Should I go ahead and replace `comint-send-string` everywhere?

angerman:
@chrisdone thoughts?

chrisdone:
Just a bit busy to try this at the moment but the `make-comint-in-buffer` change looks like a good idea. It may even fix someone else's issue here: #194 Looking forward to giving it a try!

angerman:
Alright, I'll likely go ahead and adjust the `comint` calls then. And see if I can a) reproduce #194 and b) if it's resolved by this as well.

chrisdone:
I'll try this out locally meanwhile.

chrisdone:
Okay, I tried it out locally. Looks pretty good.

Most work, but the last one in the list here doesn't:

```
/Users/chris/Work/haskell-lang/src/Foo.hs:6:1: Warning:
    The import of ‘Data.List’ is redundant
      except perhaps to import instances from ‘Data.List’
    To import instances alone, use: import Data.List()

/Users/chris/Work/haskell-lang/src/Foo.hs:7:1: Warning:
    The import of ‘Data.Ord’ is redundant
      except perhaps to import instances from ‘Data.Ord’
    To import instances alone, use: import Data.Ord()

/Users/chris/Work/haskell-lang/src/Foo.hs:10:1: Warning:
    Top-level binding with no type signature: main :: IO ()
```

Not sure why, but I can probably look into that.

Let's switch to `find-file-other-window` -- as with `-frame` it pops open a new frame (OS X window) if the buffer isn't visible, whereas `-window` only switches within Emacs windows of the current frame.

angerman:
Looks like #194 actually works as expected.
intero-194

chrisdone:
> Looks like #194 actually works as expected.

Nice! 😄 👏

chrisdone:
@AlainODea looks like we'll have a fix soon ^

elisp/intero.el
```diff
@@ -1261,11 +1258,11 @@ Automatically performs initial actions in SOURCE-BUFFER, if specified."
                       (when intero-debug
                         (message "Intero arguments: %s" (combine-and-quote-strings arguments)))
                       (message "Booting up intero ...")
-                      (apply #'start-process "stack" buffer "stack" "ghci"
-                             arguments))))
+                      (get-buffer-process (apply #'make-comint-in-buffer "stack" buffer "stack" nil "ghci"
```
chrisdone: This one can stay as `start-process` because it's not actually a comint buffer, it's used in the background for the backend stuff.

chrisdone:
There's some weird thing about clicking between them that stops working. But I'll go get lunch and then come back and merge and fix any remaining bits that I see.

elisp/intero.el
angerman: Alright. In that case we probably also do not want the the `comint-simple-send` below.

angerman:
Great!

chrisdone:
I squashed it into one commit and merged it. The `:type` convention is a little weird, but I'll let it pass.

Cheers!

angerman:
👍 Well, rename to your liking. Didn't know any better :-/

chrisdone:
No worries, it's just odd. No big deal. 😉

PR #199, opened 2016-07-30
Use C-c C-z to switch to and from REPLs

purcell:
This reproduces haskell-mode's handy toggle switching and its familiar keystrokes: when switching to the intero REPL with the new C-c C-z binding, the previous buffer is noted, and then C-c C-z in the intero REPL switches back to that buffer.

The implementation here is perhaps slightly clunky, in that it adds to `intero-repl-buffer` the responsibility for optionally noting the original buffer. The alternative would be to leave that function unchanged and provide an `intero-repl-switch` wrapping function or similar. Let me know if you'd prefer this, and I'd be happy to rework the PR slightly.


chrisdone:
I would prefer it in a separate function, but I'll try out this functionality. Thanks!

purcell:
Cool, thanks. Just ping me if you want it reworking - I'm happy to do that for you.

PR #204, opened 2016-07-31
[WIP] intero-hoogle-helm

tonyday567:
... and added a touch of import management to prove concept.

See https://gist.github.com/chrisdone/78c1c1c052a853697cf3b902de086172 for how to get `stack hoogle` working.

hoogle2

### DIFF
--- a/elisp/intero.el
+++ b/elisp/intero.el
@@ -1777,6 +1777,54 @@ automatically."
                            "hoogle" "--help")
       (0 t))))
 
+(defun intero-hoogle-helm (query)
+  "Do a Hoogle search for QUERY using stack hoogle and display using helm."
+  (interactive
+   (let ((def (intero-ident-at-point)))
+     (if (and def (symbolp def)) (setq def (symbol-name def)))
+     (list (read-string (if def
+                            (format "Hoogle query (default %s): " def)
+                          "Hoogle query: ")
+                        nil nil def))))
+  (if (intero-hoogle-supported-p)
+      (progn
+        (intero-hoogle-get-worker-create)
+        (let* ((results (intero-hoogle-blocking-query query))
+               (text (cl-loop
+                      for line in (append results nil)
+                      collect
+                      (list (format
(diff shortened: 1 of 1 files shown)


chrisdone:
This is a sweet demonstration, indeed for using the `intero-hoogle-*` API I've been putting together. The API isn't finished, specifically in how it ensures the DB is created and whatnot. Before I merge or write any functionality using that, I want to make sure that it's solid. Otherwise people will be really excited to use it (rightly so!) but the setup will feel clunky.  Generally, if chaps can hold off on making use of it until that's in place that would be good.

But I really like that helm is programmable like this -- I'd been given the impression from others that it can't be used like this without taking over your Emacs. 

Relatedly, and we'll need to establish the right interaction, I'd like that if GHC gives a not in scope error, it simply automatically hoogles those symbols. If hoogle is running, it can query it very quickly for each error. So in the case of `intercalate` it could just list one auto-selected option, and `sortBy` it could list both imports unselected, and you pick one. From the UX perspective, if you use a thing, you probably know what it is that you want. Separately, querying hoogle via a type or name to see the results would be a cracking use-case for Helm, it seems.

tonyday567:
Happy for you to reopen, but I tend towards neatness in PR lists.

It was a good exploration of the space, and helm feels right for those times when your brain knows it has to do something, but it hasn't worked out what exactly.

>>>  From the UX perspective, if you use a thing, you probably know what it is that you want. 

That's one style, but sadly not one I've mastered, lol.  My starting point is often a cut paste of someone else's code snippet.  It's a big time suck for beginners having to add unfamiliar packages and imports.

The C-c C-r UX of not-quite automated but pretty fast workflow is awesome for this.

`stack hoogle` is plenty fast.  It feels like it will perform at typing speed and make a great data and doc source.

acowley:
I adore `helm`, so really liked this approach. I think it will be crucial when suggesting a term whose type fits a hole. For imports, a more restricted interface makes sense to me.

Issue #188, opened 2016-07-25
Intero-based refactorings


chrisdone:
I fear that the parsing in #184 presents a _general_ challenge for a reliable solution. Consider:

``` haskell
{-# LANGUAGE FlexibleContexts #-}
{-# OPTIONS -Wall #-}
module X where

n :: (Show (a,a),
   Num a,Show a) => a
n = 1
```

This is valid code but the regex doesn't parse it. 

For features like this it might be more reliable and generally easier to shift the refactor part to Intero which is capable of parsing it, although not entirely capable of printing reliably, we can write our own printer for class constraints. 

Strawman proposal:

``` haskell
:rewrite-at redundant-constraints 5 1 5 1
n :: (Num a) => a
```

@acowley What do you think?

acowley:
I like this a lot, not so much for this instance which is probably fixable, but as a general approach. It would be fantastic if Intero could provide the long sought after structured error messages from GHC.

acowley:
Btw, I have elisp for doing better parsing of this kind of thing that I use elsewhere, but given the pushback here and in #190, should I just hold off on fixing this until you figure out how you'd rather do this sort of thing?

chrisdone:
I'm not opposed to merging in better parsing meanwhile. Next week I'll probably be bugfixing on the repo rather than doing new features.

acowley:
I think I fixed this on my fork yesterday. When I get a chance, I'm going to take a look at that effort Steve linked to do these auto-correct things as part of `flycheck` instead so that there's a home for it as Intero goes the haskell-ide-engine route. If the `flycheck` feature pans out, it will keep the emacs parts of Intero lighter. If the `flycheck` thing is stalled, a separate minor mode for text-based auto-correct support might make sense. 

The next GHC warnings I plan to support -- import list maintenance -- will again require syntax surgery, and I'd also like to consider making some `hlint` advice actionable as I do like a few of its hints.

chrisdone:
> I think I fixed this on my fork yesterday. When I get a chance, I'm going to take a look at that effort Steve linked to do these auto-correct things as part of `flycheck` instead so that there's a home for it as Intero goes the haskell-ide-engine route. If the `flycheck` feature pans out, it will keep the emacs parts of Intero lighter. If the `flycheck` thing is stalled, a separate minor mode for text-based auto-correct support might make sense. 

I would prefer generally to lean on other externally maintained large quantities of elisp too. 

[...]

PR #209, opened 2016-08-02
Make constraint parsing robust to nested parens

acowley:
This fixes most of the issue highlighted in #188 just to avoid leaving bugs dangling.

All further correction application code is on hold until the source parsing is moved into the intero process.


chrisdone:
Cool. Cheers!

In future changes please add the `cl-` prefix to your CL functions. There's a script under `elisp` that you can run with `sh check.sh` to check it, but at the moment only I'm using that. I'll eventually add an elisp test suite that we have in the CI, though. 👍

acowley:
Oh, thanks for the tip! I was unsure about that naming.

PR #215, opened 2016-08-03
Sort on column as well.

### LINKED ISSUE #202: Quickfix does not always work correctly.
The QuickFix window (`C-c C-r`) messes up if multiple changes are selected.

I assume this is due to the quick fix suggestions being based on the original file, but
changing the buffer changes the positions. Not sure if markers could help here properly?
Maybe if you go backwards from the End of the file?

On a related note: The quick fix window could possibly group options together.
E.g. if selecting change X to Y offer to apply this change globally?

angerman:
Quickfix already sorts the lines backwards to apply fixes, as they may
change the source location.  This however bearks down if two changes
happen in the same line.  Hence we add the column to the sort criteria
as well, such that corrections in the same line are applied in reverse
as well.

Fixes #202

### DIFF
--- a/elisp/intero.el
+++ b/elisp/intero.el
@@ -1985,10 +1985,13 @@ suggestions are available."
         (message "No changes selected to apply.")
       (let ((sorted (sort to-apply
                           (lambda (lt gt)
-                            (> (or (plist-get lt :line)
-                                   0)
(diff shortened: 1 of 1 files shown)


chrisdone:
Nice, thanks! I was hoping someone would write this for me.

PR #216, opened 2016-08-03
Add program identifier to version command flag.

AndrewRademacher:
This will be used so that stack can interrogate what executable it is running against to know if it should use features available in intero, or just GHCi.


chrisdone:
This requires an update to the Emacs Lisp which uses this. I'll apply that before merging this.

Issue #225, opened 2016-08-12
update intero.el intero-package-version from 0.1.15 to 0.1.16?


juhp:
It seems `intero-package-version` defined in `intero.el` hasn't been updated to 0.1.16?  
Is that intentional or should it be updated?  (Maybe a PR would have been quicker:)

chrisdone:
Related discussion https://github.com/commercialhaskell/intero/issues/159

Are you encountering this due to a problem or because you were reviewing the code?

juhp:
I see thanks, I did try to search the issues.

I encountered this when I tried to check the installed version of intero from within Emacs. I think it is a bit confusing: might be good to add a comment to intero.el.

chrisdone:
Yeah, I'll pop a comment in the docstring for it.

chrisdone:
Added a comment.

juhp:
Thanks

PR #226, opened 2016-08-12
intero.el: bump intero-package-version to 0.1.16

juhp:


juhp:
This would fix #225

Blaisorblade:
IIUC can be closed after the resolution of #225.

Issue #227, opened 2016-08-12
Message instead of error for lack of suggestions


kerscher:
When there is a lack of suggestions for `C-c C-r`, an elisp error shoots out to the end user.
Since this absence does not necessarily imply a mistake in intero’s program flow, it seems nicer to just message instead of dumping into a debugging session.

chrisdone:
I agree.

PR #228, opened 2016-08-12
Message instead of error on lack of suggestions.

kerscher:
Very simple fix for https://github.com/commercialhaskell/intero/issues/227.
I noticed that the entire package uses this style, but in this case it does not seem like dropping into a debugger brings any advantage. It saves the user one `q` :)


chrisdone:
Applied a slightly different change.

PR #237, opened 2016-08-23
Don't drop the first company completion

acowley:
This is related to a comment I just left on an older commit; hopefully this adds some clarity.

I don't know why this `cdr` is needed. When intero gives back a single completion, this `cdr` drops it, preventing `company` from showing any completions at all. I may be unaware of the scenario where the `cdr` is needed. If so, apologies for the disturbance!

### DIFF
--- a/elisp/intero.el
+++ b/elisp/intero.el
@@ -1171,7 +1171,7 @@ passed to CONT in SOURCE-BUFFER."
         (mapcar
          (lambda (x)
            (replace-regexp-in-string "\\\"" "" x))
-         (cdr (split-string reply "\n" t))))))))
+         (split-string reply "\n" t)))))))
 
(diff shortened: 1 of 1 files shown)


chrisdone:
I just tested this too after reading your commit. I think you caught the strange completion bug!

The `cdr` on the `repl` completion is because it returns: `("3 3 " "putChar" "putStr" "putStrLn")`

But yeah, I probably copy-pasted the repl function for the `complete-at` which I implemented later. 

Good job! :clap: 😄

acowley:
It was a funny bug hunt: I spent 90% of it with the notion it was due to some emacs weirdness like dynamic scope somewhere. I traced all through the `:async` `company` parts, reading how support for that was added to `company`, poking at the closure used for the callback, etc. Then it was just a flippin' `cdr` sitting right there 😱

Issue #242, opened 2016-08-26
Intero automatically saves buffer when flychecking, which breaks other modes.


ntc2:
The `intero-check` calls `intero-save-silently` to save the buffer. This leads to automatic buffer saves in the default flycheck configuration where `idle-change` and `new-line` are part of `flycheck-check-syntax-automatically`. This automatic saving causes at least two problems:

1. other modes that are triggered by saves don't work correctly. E.g., this breaks [GitGutter](https://github.com/syohex/emacs-git-gutter). Note that `C-x C-s` runs `save-buffer` which is "smart" and only saves the buffer if modified, so you can't simply do an extra `C-x C-s` to work around this.

2. you lose the visual indication of whether you have any unsaved changes or not. Not really sure why this is a big deal, but I'm used to having this feedback since forever in Emacs.

Related, closed issues: #148, #196.

Work around: put `(setq flycheck-check-syntax-automatically '(mode-enabled save))` in your `~/.emacs`. For larger projects, also doing `(setq haskell-tags-on-save nil)` is useful, since that's set by default in Intero, but isn't triggered by `intero-save-silently`, so you don't notice how slow it is until you disable idle Flychecking.

mgsloan:
Hey Nathan!  Thanks for looking into the details of what's going on there.  I've also seen this mess up magit, such that after checkout, I still have the old buffer state, or something along those lines.

chrisdone:
The file is actually saved, so you don't have any unsaved changes when using intero with flycheck.

The problematic part of the current implementation is that in order to save the buffer while editing, we need to disable save hooks, otherwise it'll run all sorts of hooks while you're working.

The underlying problem is that making a copy of the file and checking that would require file path munging in various places, but it looks like it may be the only solid solution.

purcell:
Yes, big +1 for this issue -- the current implementation breaks a number of things for me, from before-save hooks (stylish, hindent) to the simple expectation that I can control when and if to save the file.

I'm happy to help with the temp-file-copying solution, because I agree that it is the only viable solution.

purcell:
My pull request #248 eliminates the silent saving, and therefore this issue.

PR #248, opened 2016-09-04
Write buffer to a temp file rather than silently forcing save

purcell:
This is an initial solution to #242, wherein `intero`'s magical silent saving of buffers was causing all sorts of problems.

The implementation here uses a buffer-local variable to track an associated temp file, the contents of which will be kept up-to-date with the buffer contents. Intero backend calls are then made against that buffer, and the flycheck results translated back accordingly.

With these changes, all save-related behaviour (including before/after save hooks) works as expected.

AFAICT this all works fine. The implementation is slightly inefficient, in that it will sometimes lead to the buffer being redundantly re-written to the temp file, but it's fast and straightforward to understand. If you'd prefer, I'd be happy to work on streamlining it.

### DIFF
--- a/elisp/intero.el
+++ b/elisp/intero.el
@@ -118,14 +118,13 @@ This causes it to skip building the target."
     (when (fboundp 'interactive-haskell-mode)
       (message "Disabling interactive-haskell-mode ...")
       (interactive-haskell-mode -1)))
-  (when (intero-buffer-file-name)
-    (if intero-mode
-        (progn (flycheck-select-checker 'intero)
-               (flycheck-mode)
-               (add-to-list (make-local-variable 'company-backends) 'company-intero)
-               (company-mode)
-               (setq-local eldoc-documentation-function 'eldoc-intero))
-      (message "Intero mode disabled."))))
+  (if intero-mode
+      (progn (flycheck-select-checker 'intero)
+             (flycheck-mode)
+             (add-to-list (make-local-variable 'company-backends) 'company-intero)
+             (company-mode)
+             (setq-local eldoc-documentation-function 'eldoc-intero))
+    (message "Intero mode disabled.")))
 
 (define-key intero-mode-map (kbd "C-c C-t") 'intero-type-at)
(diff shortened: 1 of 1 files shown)


elisp/intero.el
```diff
@@ -490,10 +458,9 @@ CHECKER and BUFFER are added to each item parsed from STRING."
                        line column type
                        msg
                        :checker checker
-                       :buffer (when (string= (intero-buffer-file-name buffer)
-                                              file)
+                       :buffer (when (string= temp-file file)
                                  buffer)
-                       :filename file)
+                       :filename (intero-buffer-file-name buffer))
```
chrisdone: We may want to also do a search-and-replace for the temp filename in error messages, which GHC sometimes references.

chrisdone:
Interesting approach. I like it.

purcell:
Thanks for merging. If any issues arise from this, just @ tag me in, and I'll be happy to help.

chrisdone:
Just tried it out, here's an example of the above:

``` haskell
 Par...   68  28 error           Couldn't match expected type ‘[t0] -> P (a SrcSpanInfo)’
                with actual type ‘P (a SrcSpanInfo)’
    Relevant bindings include
      p :: P (a SrcSpanInfo)
        (bound at /var/folders/2s/6j1qqrgd57bg9ctkltrkn3z80000gn/T/intero52962niH.hs:68:22)
      normalParserNoFixity :: P (a SrcSpanInfo)
                              -> Maybe [Fixity] -> P (a SrcSpanInfo)
        (bound at /var/folders/2s/6j1qqrgd57bg9ctkltrkn3z80000gn/T/intero52962niH.hs:68:1)
    The function ‘p’ is applied to one argument,
    but its type ‘P (a SrcSpanInfo)’ has none
    In the expression: p []
    In an equation for ‘normalParserNoFixity’:
        normalParserNoFixity p _ = p []... (intero)

```

If we pop in a search-and-replace for that filename, it would mitigate the noise, I think.

shanemikel:
@purcell, do you mind taking a look at #279?

PR #257, opened 2016-09-13
Support completion of qualified identifiers

### LINKED ISSUE #66: Completion does not work for qualified modules
With this code company shows `forM` `forM_` and `forever` as posible completions

```haskell
module Main where

import Control.Monad

main :: IO ()
main = do
  for|--pointer is here
  putStrLn "hello"
```

But if code is written like this, nothing is shown as completion

```haskell
module Main where

import qualified Control.Monad as CM

main :: IO ()
main = do
  CM.for|--pointer is here
  putStrLn "hello"
```

Expected: `CM.forM` `CM.forM_` and `CM.forever` are shown as posible completions

Thanks in advance!

acowley:
Fixes #66

### DIFF
--- a/src/GhciFind.hs
+++ b/src/GhciFind.hs
@@ -13,6 +13,9 @@ module GhciFind
 import Module
 #endif
 import           Control.Exception
+#if __GLASGOW_HASKELL__ < 710
+import           Data.Foldable (foldMap)
+#endif
 import           Data.List
 import           Data.Map (Map)
 import qualified Data.Map as M
@@ -27,6 +30,27 @@ import           Name
 import           SrcLoc
 import           System.Directory
 import           Var
+
+-- | Check if there is are imported modules that should be searched
+-- for the completion sample string. If so, returns the name qualifier
+-- (i.e. module name or alias), the identifier prefix to search for,
+-- and the `ModuleName`s of the modules in which to search.
+findQualifiedSource :: [ImportDecl n] -> String
+                    -> Maybe (String, String, [ModuleName])
+findQualifiedSource importDecls sample =
+  do (ident,qual) <- breakQual sample
+     mnames <- (\nms -> if null nms then Nothing else Just nms)
+                 (foldMap (maybeToList . knownAs qual) importDecls)
+     return (qual++".", ident, mnames)
+  where breakQual xs = case break (== '.') (reverse xs) of
+                         (h,_:t) -> Just (reverse h, reverse t)
+                         _ -> Nothing
+        knownAs qual m
+          | qual == moduleNameString name || maybe False (qual ==) (asName m) =
+            Just name
+          | otherwise = Nothing
+          where name = unLoc (ideclName m)
+                asName = fmap moduleNameString . ideclAs
 
 -- | Find completions for the sample, context given by the location.
 findCompletions :: (GhcMonad m)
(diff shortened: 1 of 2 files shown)


acowley:
It turns out this is much more complicated in GHC 8.0.1 than in 7.10.3. It'll need a bit more TLC.

acowley:
I enabled Travis on my fork since something was hinky with the builds here, and they're all green there at this point. Hopefully they go green here, too, soon.

acowley:
The GHC8 support is ugly: In GHC8 , the `ModSummary` data type dropped the `ImportDecl`s we need to navigate qualified names. To regain this functionality, I had to parse the import statements from the source file myself using the starting line numbers of the `Located` imported `ModuleName`s. This is possibly tolerable when editing a file, but getting it to work at the REPL seems like it would require another approach. I haven't looked into it yet.

Even with regular files, there may be issues. For instance, literate Haskell files probably won't work properly due to the way the raw text of the import statements is extracted from the source file.

On the other hand, getting completions when using the usual one letter `as` imports surely does feel nice.

chrisdone:
Nice work! It's rather a pain in the arse that they had to remove that field. 😭 

I just pushed [this commit](https://github.com/commercialhaskell/intero/commit/fcac7d023c123f8b510b64bd68d2201cb1162493) which gives you a `[LImportDecl Name]` -- perhaps that would make implementing this easier? [ImportDecl](https://downloads.haskell.org/~ghc/7.10.3/docs/html/libraries/ghc-7.10.3/HsImpExp.html#t:ImportDecl) has the `ideclAs :: Maybe ModuleName` field you need. :metal:

src/GhciFind.hs
```diff
@@ -49,12 +83,30 @@ findCompletions infos fp sample sl sc el ec =
...
+                                   minfos <- fmap catMaybes
+                                                  (mapM getModInfo qualModNames)
+                                   if null minfos
+                                      then return noQual
+                                      else return (qual, ident, minfos)
+                                 Nothing -> return noQual
+                      else return noQual
+                let toplevelNames = concat (mapMaybe modInfoTopLevelScope minfs)
                     filteredToplevels =
-                      filter (isPrefixOf sample)
```
chrisdone: I think with `ideclAs :: Maybe ModuleName` from `ImportDecl` we can just change this to a `mapMaybe matchPrefix` where `matchPrefix :: [ImportDecl] -> String -> Name -> Maybe String` which would yield e.g. `matchPrefix .. "M.ins" "insertWith"` -> `Just "M.insertWith"` and `matchPrefix .. "sor" "sortBy" -> Just "sortBy"`.

chrisdone: Forgot to mention, `nameModule_maybe :: Name -> Maybe ModuleName` would give you the ability to match against the `ImportDecl`.

acowley: These are good ideas. I think we should also keep a very close eye on performance as this can be a very big constant factor on completion generation time in the best case. I think pre-filtering possible module sources as is done helps, but it would be great if we had a way of measuring things in a somewhat reliable way. We could put a timer on the elisp side and log that when in a debug mode.

acowley:
I rebased onto master, which simplifies this change.

acowley:
I've been successfully using this for the past couple days, so it seems like it's ready to merge.

chrisdone:
@acowley okay, let's merge.

PR #266, opened 2016-09-19
do eldoc asynchronously (fix #265)

### LINKED ISSUE #265: possible to do eldoc async?
If I profile simply moving around the code (in melpa's intero 20160907.748) , I get some "hangs"; profiler-report shows most of the cpu being spent in intero-blocking-call:

```
- timer-event-handler                                            3337  61%
 - apply                                                         3337  61%
  - #<compiled 0x28b879>                                         3059  56%
   - eldoc-print-current-symbol-info                             3059  56%
    - eldoc-intero                                               3043  56%
     - apply                                                     3008  55%
      - intero-get-type-at                                       3008  55%
       - intero-blocking-call                                    2878  53%
        - sleep-for                                               206   3%
```

Can eldoc do its thing asynchronously instead?

unhammer:
also keep around a cache to make eldoc messages show immediately where
possible, but always make a request and update the cache since we want
the Right answer as soon as we can have it without adding any
invalid-cache-complications

### DIFF
--- a/elisp/intero.el
+++ b/elisp/intero.el
@@ -719,14 +719,17 @@ instead of using `eldoc-documentation-function'."
   "ELDoc backend for intero."
   (apply #'intero-get-type-at-async
          (lambda (beg end ty)
-           (unless (string-match "^<.+>:.+:" ty)
-             (let ((msg (intero-fontify-expression
-                         (replace-regexp-in-string "[ \n]+" " " ty))))
-               ;; Got an updated type-at-point, cache and print now:
-               (puthash (list beg end)
-                        msg
-                        eldoc-intero-cache)
-               (eldoc-intero-maybe-print msg))))
+           (let ((response-status (haskell-utils-repl-response-error-status ty)))
+             (if (eq 'no-error response-status)
+               (let ((msg (intero-fontify-expression
+                           (replace-regexp-in-string "[ \n]+" " " ty))))
+                 ;; Got an updated type-at-point, cache and print now:
+                 (puthash (list beg end)
+                          msg
+                          eldoc-intero-cache)
+                 (eldoc-intero-maybe-print msg))
+               ;; But if we're seeing errors, invalidate cache-at-point:
+               (remhash (list beg end) eldoc-intero-cache))))
          (intero-thing-at-point))
   ;; If we have something cached at point, print that first:
   (gethash (intero-thing-at-point) eldoc-intero-cache))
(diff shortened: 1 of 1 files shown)


WraithM:
I'm trying this out right now. Seems great so far 👍

WraithM:
After running for a while, I get this in the messages buffer:
```
eldoc error: (wrong-type-argument wholenump nil)
Error during redisplay: (ucs-normalize-hfs-nfd-post-read-conversion 14) signaled (error "Variable binding depth exceeds max-specpdl-size")
eldoc error: (wrong-type-argument wholenump nil)
eldoc error: (error Lisp nesting exceeds ‘max-lisp-eval-depth’) [20 times]
Error while checking syntax automatically: (error "Lisp nesting exceeds ‘max-lisp-eval-depth’")
```

unhammer:
@WraithM , I haven't seen that, does it happen on any open source project I could test? 

I've only tested on Emacs 25.1.50.2 (2:201609150251~ubuntu16.04.1 from http://ppa.launchpad.net/ubuntu-elisp/ppa/ubuntu ), if that matters.

unhammer:
@WraithM I saw a possible problem in the code; could you try again and see if you can still reproduce?

WraithM:
Okay, I pulled your change and tried to trigger the issue again. Everything works great now! As far as I can tell, you fixed the issue.

Unfortunately, I'm working on a closed source codebase. Perhaps this is relevant: my issue occurred in the test-suite. I did `M-x intero-targets RET --test RET` and switched to the test-suite. I'm on Emacs 25.1.1 on OS X El Capitan installed from brew, and I use evil-mode.

acowley:
I cherry picked these commits onto my qualified completions PR and they work great together. I haven't done long-term testing, but this moves eldoc from effectively unusable for me (due to lag) to pretty nice. The main downside I see has nothing to do with this improvement, but the types returned can be incorrect when the file isn't type checking.

unhammer:
@WraithM I'll try some target-switching then :)
@acowley good to hear =D

It looks like Travis failed downloading something: https://travis-ci.org/commercialhaskell/intero/jobs/161553437 – could someone with access give it a restart?

unhammer:
Can I use `haskell-utils-repl-response-error-status` like I did in 1380cc3 or should I refrain form using functions defined in haskell-mode?

chrisdone:
Looks very nicely put together. Merging!

unhammer:
Thanks! By the way, I noticed a few other REPL-responses that probably shouldn't be given eldoc messages; but maybe that stuff should go in haskell-mode? Ie.:
https://github.com/commercialhaskell/intero/compare/master...unhammer:async-eldoc?expand=1

PR #272, opened 2016-09-26
Be clear that warnings come from intero

jml:
I got quite confused by this warning this morning:

```
Warning (emacs): Couldn’t get the Stack project root.

This can be caused by a syntax error in your stack.yaml file. Check that out.

Otherwise, please report this as a bug!
```

Where should I report this bug? Googling the error message found me https://hackage.haskell.org/package/intero-0.1.17/src/elisp/intero.el, which lead me here.

It seems a little odd to suggest that people report a bug without giving them a hint as to where (at the time, I had thought I had disabled intero, because it clearly didn't like me trying to edit a Haskell file that wasn't in a project.)

In any case, this PR updates the warning to say it's from intero, not emacs.

### DIFF
--- a/elisp/intero.el
+++ b/elisp/intero.el
@@ -1497,7 +1497,7 @@ You can always run M-x intero-restart to make it try again.
                          (funcall func state string))
                        (setq repeat t))
               (when intero-debug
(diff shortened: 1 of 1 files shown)


chrisdone:
Cool!

PR #278, opened 2016-10-01
Replace format-message with format

k0ral:
As discussed in https://github.com/commercialhaskell/intero/commit/bb2012df5e42ce657b51cf66d88f069563c7e5d7 .
My level of confidence is low: I have never written lisp code, I simply stumbled upon [a similar issue][1] that says `format-message` should be replaced with `format`. Please double-check.

[1]: https://gitlab.com/python-mode-devs/python-mode/issues/21


chrisdone:
:+1: cheers!

PR #283, opened 2016-10-10
Display λ in prompt

zjhmale:
i can not find any reason why we don't display `λ` in prompt but `‽`. i mean in every prompt of a FP language, there will be a `λ`. 😅 

Btw, there is one thing confusing me a lot while walking through the code base, why we need to [a temp file](https://github.com/commercialhaskell/intero/blob/master/elisp/intero.el#L1048-L1057), and all the infos are coming from the temp file, e.g. [type-at](https://github.com/commercialhaskell/intero/blob/master/elisp/intero.el#L1111) ?

### DIFF
--- a/elisp/intero.el
+++ b/elisp/intero.el
@@ -934,22 +934,13 @@ changes in the BACKEND-BUFFER."
       (setq intero-repl-no-load (not (member "load-all" new-options)))
       (setq intero-repl-no-build (not (member "build-first" new-options))))))
 
-;; For live migration, remove later
-(font-lock-remove-keywords
- 'intero-repl-mode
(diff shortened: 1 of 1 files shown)


Blaisorblade:
> Btw, there is one thing confusing me a lot while walking through the code base, why we need to a temp file, and all the infos are coming from the temp file, e.g. type-at ?

That's orthogonal, but could that be to prevent races between the user changing code and the typechecking process?

zjhmale:
@Blaisorblade Preventing races sounds reasonable :) Thx! 
>That's orthogonal, but could that be to prevent races between the user changing code and the typechecking process?

Issue #279, opened 2016-10-01
Not working for docker stack projects


houli:
I have a project using the "basic" stack template. When I build it with

``` yaml
docker:
  enable: true
```

in my `stack.yaml` file intero stops working. Turning off docker support, deleting `.stack-work` and rebuilding and intero starts working again.
This is using the latest Docker for Mac with intero 0.1.18 for reference.

houli:
Here's some logs from the docker containers when intero is booted up.
```
Configuring GHCi with the following packages: distributed-systems
Intero 0.1.18 (GHC 7.10.3)
Type :intro and press enter for an introduction of the standard commands.
2016-10-02T11:29:56.646518707Z 
2016-10-02T11:29:57.800842691Z 
<no location info>:
    can't find file: /var/folders/y1/_sd9mpx526j25qk9546fdz_c0000gn/T/intero14547Icw.hs
> > Failed, modules loaded: none.
2016-10-02T11:30:00.031501949Z 
<no location info>:
    can't find file: /var/folders/y1/_sd9mpx526j25qk9546fdz_c0000gn/T/intero14547Icw.hs
Failed, modules loaded: none.
Couldn't guess that module name. Does it exist?
```

Also, every time I restart intero a new docker container is being created not sure if this is intended behaviour.

houli:
Just another update, I can confirm that these `/var/folders` paths are present on my host Mac filesystem but the Docker containers running intero are failing to find them as they don't exist within the container

chrisdone:
Does `stack ghci` work?

houli:
Yes just the basic `stack ghci` works as expected. Sorry I should clarify this is using the `simple` template not `basic` as I said before but applies to any project. I've also noticed this while using the `stack-build` container for LTS 7.3 as well as LTS 6.19 so it doesn't appear to be LTS specific.

shanemikel:
I'm having this problem too, for this file: /var/folders/xz/99z8msbn5_s_c8prs1x5h5hr0000gn/T/intero18946_Ms.hs

If it's a docker project, why should intero require anything from the host?  I would think, as stack installs itself in the container, that intero would as well.

shanemikel:
So, I believe that Stack (with that Docker flag) forwards all Stack commands (reinvokes with Docker exec) to another Stack executable in the container.

It looks like Intero (Emacs client) is writing changes to working buffer in a temporary file since #248, and clearly the container won't have this path.

[...]

PR #293, opened 2016-10-20
Store temporary working files under .stack-work, not /tmp

purcell:
See #279. When enabling docker for stack, the docker image cannot see /tmp paths on the host system. With this change, .stack-work/intero/ is used as the location for temporary copies of the current working source file.

This commit does not change the containing directory for *all* intero temp files, because it is not clearly necessary to do so, but that might 
be a logical subsequent step for the sake of consistency.

Moving temp files to .stack-work/intero/ potentially allows intero to proactively delete all its temp files upon exit, which would be a win, since the code was previously relying on the OS to tidy up the temp dir.

### DIFF
--- a/elisp/intero.el
+++ b/elisp/intero.el
@@ -1075,7 +1075,12 @@ The path returned is canonicalized and stripped of any text properties."
     (prog1
         (or intero-temp-file-name
             (setq intero-temp-file-name
-                  (intero-canonicalize-path (make-temp-file "intero" nil ".hs"))))
+                  (intero-canonicalize-path
+                   (let ((temporary-file-directory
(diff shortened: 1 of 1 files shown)


purcell:
(Build failure seems unrelated.)

chrisdone:
Cool!

chrisdone:
Relatedly I probably have to fix this: https://github.com/commercialhaskell/intero/blob/master/elisp/intero.el#L936

purcell:
> Relatedly I probably have to fix this: https://github.com/commercialhaskell/intero/blob/master/elisp/intero.el#L936

Yep, I think so.

PR #296, opened 2016-10-22
Consistently use .stack-work/intero for temp files

purcell:
This allows all temp files to be seen by the stack commands, even when stack's docker support is enabled. See #279, #293.

### DIFF
--- a/elisp/intero.el
+++ b/elisp/intero.el
@@ -933,7 +933,7 @@ If PROMPT-OPTIONS is non-nil, prompt with an options list."
     (insert (propertize
              (format "Starting:\n  stack ghci %s\n" (combine-and-quote-strings arguments))
              'face 'font-lock-comment-face))
-    (let ((script (with-current-buffer (find-file-noselect (make-temp-file "intero-script"))
+    (let ((script (with-current-buffer (find-file-noselect (intero-make-temp-file "intero-script"))
                     (insert ":set prompt \"\"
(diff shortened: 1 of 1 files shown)


purcell:
@chrisdone I wondered whether `intero-destroy` would perhaps be the right place to delete the `.stack-work/intero` directory? How would we know when all intero processes for a given stack project are torn down?

shanemikel:
Do we need to know when Intero process is dead, or is it enough to have the build attempt completed?  I suppose the reload command would require a persisting tmp source file, but would the behavior of this command meet our expectations anyhow?  If so, it needs to be updated upon a buffer change in Emacs (right?).

Further, I'm tempted to suggest lack of cleanup of this particular file will be far dwarfed by the garbage Stack leaves around regularly.

If cleanup is, however, simultaneously desirable and infeasible, we can request upstream support (Stack) of a mounted directory in /tmp...

purcell:
Yeah, I'm not personally bothered by a few temp files being left around in the stack work dir.

shanemikel:
I was merely speculating, as I have zero familiarity with the Intero code.  @chrisdone, what do you think about this?

chrisdone:
I'm not too bothered either. We could definitely do _something_, like waiting for the process to start and then deleting the files, but we have more prominent issues to tackle.

chrisdone:
Thanks @purcell!

PR #297, opened 2016-10-22
Change intero-goto-definition success return value

cydparser:
This change allows one to fall back to a different lookup if Intero does not know about the symbol. E.g. I use Intero with codex: `(or (intero-goto-definition) (find-tag (find-tag-default)))`


chrisdone:
:+1: Fair change.

Issue #311, opened 2016-11-18
Literate Haskell not recognised (intero 0.1.18, ghc 8.0.1)


patrikja:
On loading a file "Bug.lhs" with only "main = print 1738" in it intero gives me
error:
    parse error on input ‘>’
as if this had been a non-literate Haskell file.

It looks like the file that intero (ghci) actually loads is a copy of my "Bug.lhs" with .hs ending:
patrikj@jansson:~/bug$ more .stack-work/intero/intero20874i_l.hs 
> main = print 1738

Is there an easy workaround?

/Patrik

To reproduce:

echo "> main = print 1738" > Bug.lhs
emacs Bug.lhs

Context: After installing fresh stack and adding this to .emacs.d/init.el
(require 'package)
(add-to-list
  'package-archives
  '("melpa" . "http://melpa.org/packages/") t)
(package-initialize)
(package-refresh-contents)

;; Install Intero
(package-install 'intero)
(add-hook 'haskell-mode-hook 'intero-mode)

and using 
resolver: lts-7.9

patrikj@jansson:~/bug$ stack --version
Version 1.2.0, Git revision 241cd07d576d9c0c0e712e83d947e3dd64541c42 (4054 commits) x86_64 hpack-0.14.0
patrikj@jansson:~/bug$ stack exec -- ghc --version
The Glorious Glasgow Haskell Compilation System, version 8.0.1
patrikj@jansson:~/bug$ more stack.yaml 
resolver: lts-7.9

chrisdone:
Correctly spotted. The extension should be taken from the original file, not assumed .hs.

chrisdone:
Pinging @purcell 

https://github.com/commercialhaskell/intero/blob/master/elisp/intero.el#L1094

Any chance you can tackle this? I've got quite a backlog of intero issues to go through.

purcell:
Sure. #321 should do the trick.

chrisdone:
@patrikja This was fixed in #321 by @purcell. Can you confirm?

patrikja:
Thanks fort the quick action.

I'd be happy to confirm, but how do I go about testing it? (How can I tell
emacs to use an intero version from github?)

/P

chrisdone:
If you're using MELPA you can just wait for it to update and reinstall the package, I think.

patrikja:
I confirm the the bug has been fixed (melpa: intero 20161123.715).
Thanks!
/P

purcell:
Yay!

PR #321, opened 2016-11-23
Use .lhs for temp files when necessary to match the current buffer

purcell:
This addresses #311, in which .hs was wrongly used unconditionally.


chrisdone:
Good call with the conditional. Thanks @purcell!

purcell:
Happy to help. If you want to add me as a collaborator for the ease of making occasional little fixes like this to the elisp side of things, that might work.

chrisdone:
@purcell Done!

purcell:
Cool, thanks.

Issue #303, opened 2016-11-04
Work under tramp


k-bx:
This is definitely an "I wish one day" item, but it'd be nice if intero would be able to launch itself on a remote machine when you open a remote file. That would let you easily program from a weak machine over ssh to a "beefy" server. Text-mode emacs could do that, of course, but tramp would be also nice, would use your local config so you wouldn't have to put it on server plus things like scrolling would still work.

WraithM:
That would be amazing. I also sometimes use tramp for accessing code on a locally hosted virtual machine, eg. Linux virtual machine on a macOS host. It would be nice if intero worked there.

chrisdone:
I think the diff to get from here to there wouldn't be too large. I'd also like this once in a while.

rlupton20:
I've been using this for the last couple of months successfully. Are there any further tweaks that need doing here? I'd be happy to take a look if there are, otherwise, this can be closed?

k-bx:
Great news! I'll check it out, but I think it makes sense closing this anyways.

PR #340, opened 2016-12-27
Made basic features work over TRAMP

rlupton20:
I've been working on making intero work over TRAMP, and this is a possible first chunk of work in that direction. Editing over TRAMP is a regular use case for me, and I discovered intero a few days ago - it looks good! Partial resolution of #303 . Local intero works as before (to the best of my knowledge). Over TRAMP, intero with these changes will

* Install intero in the stack environment (if needed)
* Run an intero background process
* Load files into an intero REPL (including loading the "prompt change script")
* Perform goto-definition
* Do type at point

company and flycheck integrations don't appear to work at the moment (next things on my list), and everything else goes as untested (and largely unmodified). I thought it best to get feedback on the work so far, for improvements/reworks or for suggestions of a better approach. I've changed calls to create background processes with TRAMP compatibile ones, and then worked through some features to get them to work as expected.

Some details
==========

I pulled out `intero-call-process` as a TRAMP-safe `call-process`. Replaced all instances of `call-process` with this. At the moment it is just another name for `process-file`, but it might be expanded on later to e.g. provide `nix` compatibility (or not, but it allows users to override it).

The main complications arises from the fact that emacs needs a TRAMPified path for files, but the intero background process needs a local path. Functions using file paths often need modifying to clarify which kind of file path is needed in each context. I opted to make the TRAMPified path the default for the temporary file buffer-variables, and to provide functions to localise or extend paths by the host they are on, and modify each file path in each context. I couldn't see a nicer way to manage this.

Initially the buffer-file-name provided this information, but the REPL for instance doesn't have a backing file, so I created a new buffer-local variable for the host of the buffers. The REPL buffer has this explicitly set from the buffer it is launched from. Technically this variable is neither "intero global" nor buffer-local - its "project global" - I find making it buffer-local dissatisfactory but I couldn't see a better way - thoughts welcome. 

Commit notes:
===========

Made type-at-point, goto definition, and repl work over TRAMP.

intero is installed, and process launched on remote machine. Other features untested at present.

Squash of the following:

Factored out call-process as separate function. Trampified process launcher.

call-process replaced by intero-call-process, which at the moment is just a wrapper for process-file (essentially a tramp aware version of call-process). start-process was replaced by start-process-file which is also tramp aware.

Made intero-make-temp-file work with tramp

intero-make-temp-file now makes temporary files on remote machines if the current buffer contains a file from a remote machine. Functions which use the temporary file don't yet cope with the remote path.

Wrote a localization function for paths

Wrote a function to localize a path to a remote machine (or return the path if not remote). This can be used to by intero to load files on a remote box, but keep the remote part in the temp filename path.

Adjust users of temporary file to relativise for host

The hostname of the temporary file is removed when it used on the remote box, and the host is added when a remote file is referenced from the local machine.

Factored out adding remote to paths

Factored out a small pattern of getting the current buffer's host and pasting in on the front of a path if it is remote, in such a way that emacs will know to feed it to TRAMP.

Got the REPL to work over TRAMP

The REPL buffer needs to know about the host for the project, but it doesn't have a file backing it, so it can't deduce the host from that. I introduced a new buffer-local variable for the host of a buffer, and rewrote the buffer host discovery functions to set this. The REPL buffer has it's buffer-local variable explicitly set, by copying the value from the buffer from which it was created.

The REPL buffer needs this information so it can cd to the correct remote path, and create temporary files in the correct place.

Small tidy

Remove a redundant function. Small simplifications + fixes.

Fixing REPL details

intero-repl-mode cleared buffer local variables, so set the buffer-host variable after intero-repl-mode is called. Wrapped the path sent to cd instead.

### DIFF
--- a/elisp/intero.el
+++ b/elisp/intero.el
@@ -350,9 +350,10 @@ Returns nil when unable to find definition."
           (xref-push-marker-stack)
         (with-no-warnings
           (ring-insert find-tag-marker-ring (point-marker))))
-      (let ((file (match-string 1 result))
-            (line (string-to-number (match-string 2 result)))
-            (col (string-to-number (match-string 3 result))))
+      (let* ((returned-file (match-string 1 result))
+	     (line (string-to-number (match-string 2 result)))
+	     (col (string-to-number (match-string 3 result)))
+	     (file (intero-extend-path-by-buffer-host returned-file)))
         (unless (string= file (intero-temp-file-name))
           (find-file file))
         (pop-mark)
@@ -453,7 +454,7 @@ running context across :load/:reloads in Intero."
     (let ((file-buffer (current-buffer)))
       (intero-async-call
        'backend
-       (concat ":l " (intero-temp-file-name))
+       (concat ":l " (intero-localize-path (intero-temp-file-name)))
        (list :cont cont
              :file-buffer file-buffer
              :checker checker)
@@ -853,7 +854,7 @@ This is set by `intero-repl-buffer', and should otherwise be nil.")
 If PROMPT-OPTIONS is non-nil, prompt with an options list."
   (interactive "P")
   (save-buffer)
-  (let ((file (intero-temp-file-name))
+  (let ((file (intero-localize-path (intero-temp-file-name)))
         (repl-buffer (intero-repl-buffer prompt-options t)))
     (with-current-buffer repl-buffer
       (comint-simple-send
@@ -903,8 +904,14 @@ STORE-PREVIOUS is non-nil, note the caller's buffer in
             (get-buffer name)
           (with-current-buffer
               (get-buffer-create name)
-            (cd root)
-            (intero-repl-mode)
+	    ;; The new buffer doesn't know if the initial buffer was hosted
+	    ;; remotely or not, so we need to extend by the host of the
+	    ;; initial buffer to cd. We could also achieve this by setting the
+	    ;; buffer's intero-buffer-host, but intero-repl-mode wipes this, so
+	    ;; we defer setting that until after.
+            (cd (intero-extend-path-by-buffer-host root initial-buffer))
+            (intero-repl-mode) ; wipes buffer-local variables
+	    (setq intero-buffer-host (intero-buffer-host initial-buffer))
             (intero-repl-mode-start backend-buffer
                                     (buffer-local-value 'intero-targets backend-buffer)
                                     prompt-options)
@@ -1025,7 +1032,7 @@ If PROMPT-OPTIONS is non-nil, prompt with an options list."
               (current-buffer)))
            (script
             (with-current-buffer script-buffer
-              (intero-buffer-file-name))))
+              (intero-localize-path (intero-buffer-file-name)))))
       (let ((process
              (get-buffer-process
               (apply #'make-comint-in-buffer "intero" (current-buffer) "stack" nil "ghci"
@@ -1245,6 +1252,31 @@ The path returned is canonicalized and stripped of any text properties."
     (when name
       (intero-canonicalize-path (substring-no-properties name)))))
 
+(defvar-local intero-buffer-host nil
+  "The hostname of the box hosting the intero process for the current buffer.")
+
+(defun intero-buffer-host (&optional buffer)
+  "Get the hostname of the box hosting the file behind the buffer."
+  (with-current-buffer (or buffer (current-buffer))
+    (let ((file (intero-buffer-file-name)))
+    (if intero-buffer-host
+	intero-buffer-host
+      (setq intero-buffer-host
+	    (when file
+	      (if (eq 0 (string-match tramp-file-name-regexp file))
+		  (tramp-file-name-host (tramp-dissect-file-name file))
+		"")))))))
+	
+(defun intero-extend-path-by-buffer-host (path &optional buffer)
+  "Take a path, and extend it by the host of the provided buffer (default to current buffer). Return path unchanged if the file is local, or the buffer has no host."
+  (with-current-buffer (or buffer (current-buffer))
+    (if (or (eq nil (intero-buffer-host)) (eq "" (intero-buffer-host)))
+	path
+      (concat "/"
+	      (intero-buffer-host)
+	      ":"
+	      path))))
+
 (defvar-local intero-temp-file-name nil
   "The name of a temporary file to which the current buffer's content is copied.")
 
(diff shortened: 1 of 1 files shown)


chrisdone:
A quick review over the code, it looks quite reasonable. I'll go through it in detail tomorrow and give feedback on anything that might be concerning.

chrisdone:
A quick review over the code, it looks quite reasonable. I'll go through it in detail tomorrow and give feedback on anything that might be concerning.

elisp/intero.el
```diff
@@ -1245,6 +1252,31 @@ The path returned is canonicalized and stripped of any text properties."
...
+(defun intero-extend-path-by-buffer-host (path &optional buffer)
+  "Take a path, and extend it by the host of the provided buffer (default to current buffer). Return path unchanged if the file is local, or the buffer has no host."
+  (with-current-buffer (or buffer (current-buffer))
+    (if (or (eq nil (intero-buffer-host)) (eq "" (intero-buffer-host)))
+	path
+      (concat "/"
+	      (intero-buffer-host)
+	      ":"
+	      path))))
+
```
chrisdone: The above changes look fine.

elisp/intero.el
```diff
@@ -350,9 +350,10 @@ Returns nil when unable to find definition."
           (xref-push-marker-stack)
         (with-no-warnings
           (ring-insert find-tag-marker-ring (point-marker))))
-      (let ((file (match-string 1 result))
-            (line (string-to-number (match-string 2 result)))
-            (col (string-to-number (match-string 3 result))))
+      (let* ((returned-file (match-string 1 result))
+	     (line (string-to-number (match-string 2 result)))
+	     (col (string-to-number (match-string 3 result)))
+	     (file (intero-extend-path-by-buffer-host returned-file)))
```
chrisdone: So this maps from `/foo/bar.hs` to `/ssh:blah.com/foo/bar/.hs`, right? Because the `intero` piped process returns a local path, whereas we want a remote-path, in order to open the right TRAMP buffer.

elisp/intero.el
```diff
@@ -453,7 +454,7 @@ running context across :load/:reloads in Intero."
     (let ((file-buffer (current-buffer)))
       (intero-async-call
        'backend
-       (concat ":l " (intero-temp-file-name))
+       (concat ":l " (intero-localize-path (intero-temp-file-name)))
```
chrisdone: Makes sense, this is what I'd imagined I'd do to provide tramp integration.

elisp/intero.el
```diff
@@ -903,8 +904,14 @@ STORE-PREVIOUS is non-nil, note the caller's buffer in
...
-            (cd root)
-            (intero-repl-mode)
+	    ;; The new buffer doesn't know if the initial buffer was hosted
+	    ;; remotely or not, so we need to extend by the host of the
+	    ;; initial buffer to cd. We could also achieve this by setting the
+	    ;; buffer's intero-buffer-host, but intero-repl-mode wipes this, so
+	    ;; we defer setting that until after.
+            (cd (intero-extend-path-by-buffer-host root initial-buffer))
+            (intero-repl-mode) ; wipes buffer-local variables
+	    (setq intero-buffer-host (intero-buffer-host initial-buffer))
```
chrisdone: I suppose `intero-buffer-host` would work on the newly created repl buffer too, because you just `cd'd` to a remote path like `/ssh:foo.com/foo/`.

elisp/intero.el
```diff
@@ -1025,7 +1032,7 @@ If PROMPT-OPTIONS is non-nil, prompt with an options list."
               (current-buffer)))
            (script
             (with-current-buffer script-buffer
-              (intero-buffer-file-name))))
+              (intero-localize-path (intero-buffer-file-name)))))
```
chrisdone: Solid 👍

elisp/intero.el
```diff
@@ -1412,7 +1452,7 @@ passed to CONT in SOURCE-BUFFER."
   (intero-async-call
    'backend
    (format ":complete-at %S %d %d %d %d %S"
-           (intero-temp-file-name)
+           (intero-localize-path (intero-temp-file-name))
            (save-excursion (goto-char beg)
                            (line-number-at-pos))
            (save-excursion (goto-char beg)
```
chrisdone: All the above intero calls look legit.

elisp/intero.el
```diff
@@ -1455,6 +1495,11 @@ Completions for PREFIX are passed to CONT in SOURCE-BUFFER."
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 ;; Process communication
 
+(defun intero-call-process (program &optional infile destination display &rest args)
+  "Synchronously call a process. Same interface as call-process/process-file."
+  (let ((process-args (append (list program infile destination display) args)))
+    (apply 'process-file process-args)))
```
chrisdone: I think I know the reason for this function, it ensures a tramp process is created instead of a local one when necessary. But perhaps it should be documented. It was sensible to make an abstraction function for this. :+1:

elisp/intero.el
```diff
@@ -1791,7 +1837,7 @@ You can always run M-x intero-restart to make it try again.
 (defun intero-get-buffer-create (worker)
   "Get or create the stack buffer for WORKER.
 Uses the directory of the current buffer for context."
-  (let* ((root (intero-project-root))
+  (let* ((root (intero-extend-path-by-buffer-host (intero-project-root)))
```
chrisdone: So this will have the effect that the intero worker buffer will be created under `/ssh:foo.com/blah` when proper. That's good.

chrisdone:
I'm going to do a re-indent of the whole file, as you introduced some mis-indentations. Just remember to pull from master before you continue working, to avoid you having boring merge conflicts later.

chrisdone:
@rlupton20 I pushed three commits, make sure to pull those before you proceed!

chrisdone:
flycheck seems to work for me. Testing on a remote machine! It's not fast over SSH, but it works. Yay! 🎉

elisp/intero.el
rlupton20: Exactly as you say.

elisp/intero.el
rlupton20: I'm not sure - it calls `intero-buffer-filename` in the background, and I wouldn't have thought the working directory is enough to provide that. `intero-repl-mode` blows it all away, in any case.

elisp/intero.el
rlupton20: Exactly. The docstring is really lacking in that main point! I'll expand that in future work.

rlupton20:
Thanks, I'll pull all your changes before continuing. Sorry about the indentation, I should really fix that; my emacs is a little out of whack at times, and I've never properly investigated why.

My flycheck doesn't seem to do anything (yet it works locally). I'll investigate it, since one of the nice things about intero is the out-of-the-box functioning goodness. Glad its working in some cases though!

PR #343, opened 2016-12-29
Made flycheck work over TRAMP

rlupton20:
Made flycheck work over TRAMP. Tidied a couple of other odds and ends from previous TRAMP work.

Key work in `intero-parse-errors-warning-splices`. `flycheck` wasn't correctly identifying the buffer to display errors in, because it was comparing a local and remote path (with `string=`). I made the paths remote (when remote), and wrote a comparison function for paths to replace `string=`, since TRAMP paths have a few variations.

Furthers solving #303. Builds on #340.

Commit notes
===========

The background process was finding errors, but the flycheck backend was not identifying the buffer to display errors in. This was because it was comparing the absolute path from the buffer, and the local path from the background process. Fixed this, and provided a general utility function for comparing paths (since paths over TRAMP may differ despite representing the same file).

Did a couple of small improvements to the TRAMP compatibility code also (docstrings, factoring out TRAMP path identification).

Squash of:

Got flycheck to work over tramp

Flycheck was finding the errors, but because it didn't recognize the temporary file from emacs' perspective and from intero's perspective as being the same, it didn't display the errors.

Improved factoring

Made a function to compare paths, rather than relying on standardized paths being used as strings.

Tidying

Refactored path comparison function, so the "representation" function could be removed completely. Improved a couple of docstrings.

### DIFF
--- a/elisp/intero.el
+++ b/elisp/intero.el
@@ -504,7 +504,8 @@ CHECKER and BUFFER are added to each item parsed from STRING."
               (concat "[\r\n]\\([A-Z]?:?[^ \r\n:][^:\n\r]+\\):\\([0-9()-:]+\\):"
                       "[ \n\r]+\\([[:unibyte:][:nonascii:]]+?\\)\n[^ ]")
               nil t 1)
-        (let* ((file (intero-canonicalize-path (match-string 1)))
+        (let* ((local-file (intero-canonicalize-path (match-string 1)))
+               (file (intero-extend-path-by-buffer-host local-file buffer))
                (location-raw (match-string 2))
                (msg (match-string 3)) ;; Replace gross bullet points.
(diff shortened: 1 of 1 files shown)


rlupton20:
company completion is actually also working for me, but only if I set `company-idle-delay` to something other than the default `0.5` (and that's only an issue over TRAMP). Fairly sure that's not an issue with intero (found this: https://groups.google.com/forum/#!topic/gnu.emacs.sources/JUcjGrLMV40), but I might have a go at tracking it down.

A temporary workaround could be to set `company-idle-delay` to `0.49`...

rlupton20:
Yes, it doesn't look like intero code is entered at all. It looks like company just doesn't trigger completion. Interesting bug!

chrisdone:
Thanks, the changes look very reasonable. Merging and I'll try it out!

Issue #310, opened 2016-11-15
Feature request: Expand & display template haskell expression


saurabhnanda:
Would it be possible to take my cursor on the following expression, hit a keybinding, and view the expanded splice:

```
$(defineTable "roles")
```

meditans:
@saurabhnanda if you do `:set -ddump-splices` in your ghci panel, you will see the generated splices on reload. It's still not comfortable as viewing only the splice, but it's a start!

saurabhnanda:
Do I have to `C-c C-l` (load the current buffer in GHCi) for that to take effect?

meditans:
yes @saurabhnanda, and the splices will be below the eventual compilation errors and above the new prompt.

PR #346, opened 2017-01-02
Template haskell expansion

rlupton20:
Wrote functions to expand template haskell splice at point. point needs to be over the template haskell expression being expanded (rather than over parentheses or `$`). Think this is reasonable. Solves #310.

GHCi can be a bit awkward about column numbering, according to whether a template haskell expression is parenthesised or not (parentheses seem to +1 everything, if the "thing at point" is a parenthesised single character, only a single column is provided, rather than start and end...). I think I have this covered sensibly in the regex that grabs the expansion.

Overview of changes
* Load intero backend with `-ddump-splices` and `-XTemplateHaskell`.
* Add function to expand template haskell at point. This works by making the intero backend load the buffer's temporary file, and then regexing out the correct expansion.
* Shortcut `C-c C-e` bound to this for testing. Didn't set this with any particular thought (except e for expand), so might be worth setting to something else if there are strong opinions on the matter.

Tested over TRAMP (but not locally) with the following splices:
```
g = $template
g' = $f
g'' = $(f)
this = Just "filling a gap"; a = $f
h = $(template); g''' = $f
k = $(template' 3)
```
where
```
f :: Q Exp
f = [| \x -> x |]

template :: Q Exp
template = [| \x ->
               case x of
                 Nothing -> 0
                 Just _ -> 1 |]

template' :: Int -> Q Exp
template' x = [| \y -> x |]
```

### DIFF
--- a/elisp/intero.el
+++ b/elisp/intero.el
@@ -384,24 +384,25 @@ Returns nil when unable to find definition."
 (defun intero-expand-splice-at-point ()
   "Show the expansion of the template haskell splice at point."
   (interactive)
-  (let* ((line (line-number-at-pos (point)))
-	 (pos (intero-thing-at-point))
-	 (beg (car pos))
-	 (start-column (save-excursion (goto-char beg)
-				       (current-column))))
-    (intero-async-call
-     'backend
-     (concat ":l " (intero-localize-path (intero-temp-file-name)))
-     (list :file-buffer (current-buffer)
-	   :line line
-	   :start start-column)
-     (lambda (state string)
-       (let* ((buffer (plist-get state :file-buffer))
-	      (line (plist-get state :line))
-	      (start (plist-get state :start))
-	      (expansion (intero-get-expansion-at-pos string buffer line start)))
-	 (when expansion
-	   (message (intero-fontify-expression expansion))))))))
+  (unless (intero-gave-up 'backend)
+    (let* ((line (line-number-at-pos (point)))
+	   (pos (intero-thing-at-point))
+	   (beg (car pos))
+	   (start-column (save-excursion (goto-char beg)
+					 (current-column))))
+      (intero-async-call
+       'backend
+       (concat ":l " (intero-localize-path (intero-temp-file-name)))
+       (list :file-buffer (current-buffer)
+	     :line line
+	     :start start-column)
+       (lambda (state string)
+	 (let* ((buffer (plist-get state :file-buffer))
+		(line (plist-get state :line))
+		(start (plist-get state :start))
+		(expansion (intero-get-expansion-at-pos string buffer line start)))
+	   (when expansion
+	     (message (intero-fontify-expression expansion))))))))
 
 (defun intero-restart ()
   "Simply restart the process with the same configuration as before."
--- a/elisp/intero.el
+++ b/elisp/intero.el
@@ -381,6 +381,14 @@ Returns nil when unable to find definition."
     (when (string-match regex string)
       (match-string 3 string))))
 
+(defmacro intero-with-dump-splices (exp)
+  "Run EXP but with dump-splices enabled in the intero backend process."
+  (list 'when (list 'intero-blocking-call ''backend ":set -ddump-splices")
+	(list 'let (list (list 'result exp))
+	      (list 'progn
+		    'nil ; Disable dump-splices here in future
+		    'result))))
+
 (defun intero-expand-splice-at-point ()
   "Show the expansion of the template haskell splice at point."
   (interactive)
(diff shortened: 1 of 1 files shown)


chrisdone:
I think this is a much-wanted feature!

elisp/intero.el
```diff
@@ -1661,6 +1699,9 @@ Automatically performs initial actions in SOURCE-BUFFER, if specified."
       (set-process-query-on-exit-flag process nil)
       (process-send-string process ":set -fobject-code\n")
       (process-send-string process ":set prompt \"\\4\"\n")
+      ;; The following settings give us expanded splices on load
+      (process-send-string process ":set -XTemplateHaskell\n")
+      (process-send-string process ":set -ddump-splices\n")
```
chrisdone: I think we should move this into `intero-expand-splice-at-point`, using a `intero-blocking-call`. I think that dump-splices significantly increases output size and in proportion reduces performance of flycheck reporting.

Because `:set -ddump-splices` can be run repeatedly to no ill effects, we should "lazily" enable it as needed. That way, if someone's doing no template-haskell development, then they don't pay the cost. If someone is working on template-haskell and they want to see the splice, they can pay for it.

Unfortunately there seems no way to turn off flags, that I can see in GHC's documentation, so once you turn it on, you can't turn it off.

chrisdone: (With some GHC API work, we could add support to turn it off explicitly. That can be something we do later.)

elisp/intero.el
```diff
@@ -366,6 +367,43 @@ Returns nil when unable to find definition."
...
+       (list :file-buffer (current-buffer)
+	     :line line
+	     :start start-column)
+       (lambda (state string)
+	 (let* ((buffer (plist-get state :file-buffer))
+		(line (plist-get state :line))
+		(start (plist-get state :start))
+		(expansion (intero-get-expansion-at-pos string buffer line start)))
+	   (when expansion
+	     (message (intero-fontify-expression expansion))))))))
```
chrisdone: Never use `message` directly on a variable, use `(message "%s" ...)`, otherwise if the variable contains `"%s"` then it breaks.

elisp/intero.el
rlupton20: Sounds reasonable. I'll make the changes.

elisp/intero.el
rlupton20: Oops! Yes makes sense, and I'm not sure I had intended to call it directly in the first place!

rlupton20: I also have a missing `)` in here :(. Will fix that too.

elisp/intero.el
rlupton20: Fixed. Had a missing `)` too. Must have accidently deleted it in the last moment.

elisp/intero.el
rlupton20: Fixed.

elisp/intero.el
rlupton20: I wrote a little macro which in future can "bracket" dump-splices for an expression, but for now just turns it on. Tested and working.

rlupton20:
Fixes:
* Bad `message` call corrected
* dump splices now enabled lazily. Stub for "bracketing" approach.
* TemplateHaskell doesn't need to be explicitly enabled, so removed

chrisdone:
Great! 👍

chrisdone:
@rlupton20 I rewrote the function so that it's less picky about which column you have to run the `C-c C-e` on. Now it works on or anywhere after the `$` in the splice expression.

rlupton20:
@chrisdone much better! Mine was definitely messy.

A couple of questions (for my own learning)
* presumably you use a blocking call because its a user request that triggers this (as opposed to flycheck which is triggered in the background, so it would be unacceptable to make the user wait for that to run each time). Is there a reason why blocking is preferable to an asynchronous call in this case / in general (for this kind of operation)?
* Were the tabbing problems caused by me again? I thought I'd dealt with that, and they didn't show up on my diffs afaik?

Will spend some time going through your rewrite. A couple of things for me to learn from there. Lack of regexs is encouraging :+1:.

chrisdone:
Blocking call because it's both user-requested and the response is expected to be immediate. Blocking preferable because you don't have to convolve your code into async style.

I think the indentation was just off or something. I run indent-region on the whole buffer regularly as a habit.

rlupton20:
Just as a further note, it also works after the splice expression. e.g. C-c C-e on the b in
`a = $(splice); b = $(anothersplice)`
expands `splice`.

Not a huge deal (only very vague possibility of confusion in e.g. `a = $(splice 3); b = $(splice 5)` type examples), but could maybe be tightened up a little. Won't worry for now, but may fix if I feel inclined one moment. Worth noting now for that day somebody gets confused :smile:.

PR #359, opened 2017-01-15
Require Emacs 24.4 rather than 24.3, due to use of string-join

purcell:
Alternate fix would be to define an "intero-string-join" locally, or just use the classic `(mapconcat 'identity strings sep)`.

### DIFF
--- a/elisp/intero.el
+++ b/elisp/intero.el
@@ -11,7 +11,7 @@
 ;; Created: 3rd June 2016
 ;; Version: 0.1.13
 ;; Keywords: haskell, tools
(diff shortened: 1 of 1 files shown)


purcell:
Incidentally, this issue (and several others) are detected by [package-lint.el](https://github.com/purcell/package-lint): it might be worth adding its batch-mode checks (and indeed all of check.sh) to the travis build to guard against breakage.

chrisdone:
@purcell I'm in favor of that. I'd like to implement a test suite for the Emacs Lisp too.

PR #373, opened 2017-02-06
* intero.el: Add intero linkify hook only locally

marsam:
dbeea0c tried to make this change, but `add-hook` local parameter is the fourth
argument not the third.


chrisdone:
Cheers!

PR #384, opened 2017-03-17
New help buffer

chrisdone:
This introduces an Intero-specific help buffer instead of re-using the emacs lisp help buffer:

1. `C-c C-i` now works inside the help buffer, so you can drill down into data types.
2. `C-c C-t` will report the type in the minibuffer as usual, but also insert it into the help buffer in the background. So if you want to drill down into a type, you can switch to that buffer and continue with `C-c C-i`.

Great for exploring a new project.

### DIFF
--- a/elisp/intero.el
+++ b/elisp/intero.el
@@ -316,11 +316,32 @@ You can use this to kill them or look inside."
 With prefix argument INSERT, inserts the type above the current
 line as a type signature."
   (interactive "P")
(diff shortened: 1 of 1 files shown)


chrisdone:
I've been using this for two months without issue, so merged in. It shouldn't cause any regressions to existing behavior. `C-c C-i` changes slightly, and `C-c C-t`'s output can now be seen in both the minibuffer and this buffer.

PR #386, opened 2017-03-18
Update check.sh to use cask and package-lint

purcell:
Further to a couple of small preemptive changes I've committed directly to `master`, I'd like to get `check.sh` running in at least one supported Emacs version as part of the travis build. On the way to that, I've overhauled the existing `check.sh` in this commit.

- Use cask to ensure required packages are installed
- Strictly require clean byte-compilation, or fail
- Fail on any package-lint errors


chrisdone:
Merging in the hopes of successful travis build.

PR #387, opened 2017-03-18
Check elisp as part of the Travis build

purcell:
This PR extends #386 by hooking `elisp/check.sh` into the Travis build. I'm filing it separately here so I can test it out. :-)


chrisdone:
What's the status of this? It seems that the build fails.

purcell:
Got that working in the end: thanks for commenting it out after merging.

PR #394, opened 2017-04-04
Allow choosing which Stack executable to use

### LINKED ISSUE #393: Using Intero in Emacs w/o Stack
Hey!

There was already a rejected attempt here →https://github.com/commercialhaskell/intero/pull/52.

Maybe we could have something simpler at first? There are only 12 hard-coded instances of the string `"stack` in `intero.el`. If I made a PR that made all 12 configurable and defaulting to their current
values, would it get accepted?

Thank you!

michalrus:
Resolves #393.

Use cases:

* testing different versions of Stack side-by-side,

* substituting `stack` with a small shim that uses a different resolver, e.g. Nix.

### DIFF
--- a/elisp/intero.el
+++ b/elisp/intero.el
@@ -124,6 +124,13 @@ To use this, use the following mode hook:
   :group 'intero
   :type 'string)
 
+(defcustom intero-stack-executable
+  "stack"
+  "Name or path to the Stack executable to use."
+  :group 'intero
+  :type 'string)
+(make-variable-buffer-local 'intero-stack-executable)
+
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 ;; Modes
 
@@ -1094,7 +1101,7 @@ If PROMPT-OPTIONS is non-nil, prompt with an options list."
               (intero-localize-path (intero-buffer-file-name)))))
       (let ((process
              (get-buffer-process
-              (apply #'make-comint-in-buffer "intero" (current-buffer) "stack" nil "ghci"
+              (apply #'make-comint-in-buffer "intero" (current-buffer) intero-stack-executable nil "ghci"
                      (append arguments
                              (list "--verbosity" "silent")
                              (list "--ghci-options"
@@ -1688,7 +1695,7 @@ Installing intero-%s automatically ...
 " intero-package-version))
       (redisplay)
       (cl-case (intero-call-process
-                "stack" nil (current-buffer) t "build"
+                intero-stack-executable nil (current-buffer) t "build"
                 (with-current-buffer buffer
                   (let* ((cabal-file (intero-cabal-find-file))
                          (package-name (intero-package-name cabal-file)))
@@ -1742,7 +1749,7 @@ Automatically performs initial actions in SOURCE-BUFFER, if specified."
                       (when intero-debug
                         (message "Intero arguments: %s" (combine-and-quote-strings arguments)))
                       (message "Booting up intero ...")
-                      (apply #'start-file-process "stack" buffer "stack"
+                      (apply #'start-file-process "stack" buffer intero-stack-executable
                              arguments))))
       (set-process-query-on-exit-flag process nil)
       (process-send-string process ":set -fobject-code\n")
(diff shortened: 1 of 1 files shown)


michalrus:
Thank you, that’s wonderful! :tada: Fixed version is now building on Travis.

(When deciding whether to make it local or not, I looked at `intero-repl-no-load` and `intero-repl-no-build` which imply various command line switches to Intero—and there’s only one Intero process per project.)

purcell:
Super, thanks - merged now.

> (When deciding whether to make it local or not, I looked at intero-repl-no-load and intero-repl-no-build which imply various command line switches to Intero—and there’s only one Intero process per project.)

Yes, I think the same argument would apply to those two vars as well, actually.

michalrus:
Maybe I can change them as well, in another PR?

There are a few more occurrences of `stack` in help/debug messages:

* https://github.com/commercialhaskell/intero/blob/dd8113bcf6d4be3db0608a5fb63bb26d6e60fffe/elisp/intero.el#L1088
* https://github.com/commercialhaskell/intero/blob/dd8113bcf6d4be3db0608a5fb63bb26d6e60fffe/elisp/intero.el#L1905
* https://github.com/commercialhaskell/intero/blob/dd8113bcf6d4be3db0608a5fb63bb26d6e60fffe/elisp/intero.el#L2015

I’m sorry I missed them—when trying to create a minimal shim for `nix-shell`, they came up… I think it would be a particularly helpful change for those most prone to breaking things, i.e. those trying to replace regular Stack with something else.

michalrus:
OK, so I wrote the Nix shim and it works: https://github.com/michalrus/kornel/commit/31fd80744d290fe7df2e88752737d7a3b020eece.

That commit is probably more/less what it takes for a project to use Intero with Nix rather than Stack → ca. 30 lines of `/bin/sh`, that are not as terrifying as I imagined they would be. :smiley: 

/cc @acowley @puffnfresh @domenkozar @kmicu

(I tested this only on NixOS, so it might not work straight away on OS X.)

Since it’s now proven that this _can_ work, maybe we could have some more integration (or decoupling from Stack)… But it might as well be too early and I don’t really have any good ideas.

domenkozar:
@shlevy

acowley:
This is excellent, thank you @michalrus! There may be minor issues with `bash`, `grep`, and `sed` on darwin unless the shim is also executed in a `nix-shell`, but I haven't tried it yet.

shlevy:
@michalrus Haven't had a chance to look at this in detail yet, but this is promising! Thanks for the ping @domenkozar

michalrus:
I slept on it and… Making each project copy-paste this shell script feels wrong.

What I would rather do is:

* define a small separate project `intero-nix-shim` (*) that would do all that argument juggling in a safer way,
* in a saner language than `sh`—I’d go with Haskell, we could even use `optparse-applicative`,
* have Nix+Intero projects depend on this executable.

Then, users would only have to add a single package to their `default.nix` _and_ a `.dir-locals.el` that would switch from Stack to this executable.

Since users would only be using this code in context of Nix(pkgs), I don’t think it would be useful to add it to Hackage/Stackage. @domenkozar, @shlevy, could I add it only to nixpkgs, then? It would save ~10 lines in `default.nix` of each project, replacing them with one word, the package name.

(*) what should be its name?

shlevy:
@michalrus Can we pair up on this? I'd love to get this working smoothly. Not sure if it will be relevant, but you may also want to check out https://stable.melpa.org/#/nix-buffer and see if it makes sense to integrate in with that.

shlevy:
OK, looking at the details here, I think this definitely fits well with `nix-buffer`, as then it's not even a per-project consideration, you just put a `dir-locals.nix` in your HOME setting `intero-stack-executable`.

shlevy:
See https://github.com/shlevy/config/blob/bb31ef5e569fb057fa17b9c7f4a28a6de4556a67/dir-locals.nix for my global `dir-locals.nix` that sets up GHC paths properly for every haskell package I work on (though obviously doesn't work for intero yet, this solves company mode and inferior ghci).

michalrus:
> @michalrus Can we pair up on this? I'd love to get this working smoothly.

:+1: I’ll be happy to! Thank you for the links and ideas, I’ll look into your `nix-buffer` soon (I hope today)! :smiley:

ljli:
I replaced (a little hamfisted) stack a while ago with nix-shell and cabal (https://github.com/ljli/intero/tree/use-nix-shell). I think you could get around using a stack shim and use cabal directly.
I want to rely more on direnv (which has integrations with nix and emacs) and was about expand the branch into proper (and configurable) support for cabal, stack, nix-shell + cabal. But I'm happy to let you guys sort it out. Do you plan for plain cabal support in the end (through a shim or not)?

michalrus:
@ljli, oh, yes, I think we’d all want native support, that would be optimal! :dreamy:

But since one attempt was already rejected (#52), I thought it could be better to try differently (at first!).

I’ll be be sure to read the diff of your branch, thank you!

PR #398, opened 2017-04-06
Don’t make intero-repl-no-load and intero-repl-no-build buffer-local

michalrus:
See @purcell’s comment here: https://github.com/commercialhaskell/intero/pull/394#issuecomment-292081836.


purcell:
Thanks!

PR #402, opened 2017-04-09
Fix propagation of buffer-local variables

### LINKED ISSUE #401: intero-stack-executable not inherited in non-file buffers
Emacs’ documentation states:

> Whenever Emacs visits any file in that directory or any of its subdirectories, it will apply the directory-local variables specified in `.dir-locals.el`, as though they had been defined as file-local variables for that file […]
> — https://www.gnu.org/software/emacs/manual/html_node/emacs/Directory-Variables.html

I missed this behavior in my initial PR (#394), because I was unknowingly setting `intero-stack-executable` _globally_ in my `.dir-locals.el` (`setq` instead of `setq-local`), which became obvious after opening a few projects at the same time.

After correcting that, despite the fact that Haskell buffers have `intero-stack-executable` changed by `.dir-locals.el`, other non-file-visiting Intero buffers do not, and, in effect, Stack is used anyway.

/cc @purcell 

I’ll try to find a way to propagate the value to these buffers, that modifies `intero.el` the least. But any suggestions would be very appreciated!

michalrus:
Resolves #401.

Every time `(with-temp-buffer)` is used, any local variables of the calling buffer are not propagated down.

One such variable is `intero-stack-executable`, which the user might want to change locally (potentially, using .dir-locals.el). Since processes are most often started in new, temporary, non-file-visiting buffers, these new buffers need to see the value requested by the user.

### DIFF
--- a/elisp/intero.el
+++ b/elisp/intero.el
@@ -263,18 +263,15 @@ This is slower, but will build required dependencies.")
 (defvar-local intero-buffer-host nil
   "The hostname of the box hosting the intero process for the current buffer.")
 
-(defmacro intero-inherit-local-variables (buffer)
+(defun intero-inherit-local-variables (buffer)
   "Make the current buffer inherit values of certain local variables from BUFFER."
-  (let ((varlist '(intero-stack-executable
-                   intero-repl-no-build
-                   intero-repl-no-load
-                   ;; TODO: shouldn’t more of the above be here?
-                   )))
-    `(progn
-       ,@(cl-mapcar
-          (lambda (symb)
-            `(setq-local ,symb (buffer-local-value (quote ,symb) ,buffer)))
-          varlist))))
+  (let ((variables '(intero-stack-executable
+                     intero-repl-no-build
+                     intero-repl-no-load
+                     ;; TODO: shouldn’t more of the above be here?
+                     )))
+    (cl-loop for v in variables do
+             (set (make-local-variable v) (buffer-local-value v buffer)))))
 
 (defmacro intero-with-temp-buffer (&rest body)
   "Run BODY in `with-temp-buffer', but inherit certain local variables from the current buffer first."
--- a/elisp/intero.el
+++ b/elisp/intero.el
@@ -253,7 +253,7 @@ This is slower, but will build required dependencies.")
 (defvar-local intero-extensions nil
   "Extensions supported by the compiler.")
 
-(defvar intero-ghc-version nil
+(defvar-local intero-ghc-version nil
   "GHC version used by the project.")
 
 (defvar-local intero-repl-last-loaded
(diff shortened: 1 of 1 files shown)


michalrus:
I’m sorry the diff is that big. :slightly_frowning_face: It also feels clunky.
On the other hand it’s mostly whitespace causing this. If only there was a contextual diffing tool…

/cc @purcell

purcell:
Thanks. A lot of this is quite repetitive, and I worry that we'll accidentally break it later.

I wonder if we should therefore be defining simple `with-intero-temp-buffer` and/or `with-intero-backend-buffer` macros which would wrap `with-temp-buffer`/`with-buffer` _and_ propagate the stack executable etc.?

michalrus:
That sounds *a lot* better! =) Thank you! I’ll try to do that tomorrow.

purcell:
> I’ll try to do that tomorrow.

Thanks. Let me know if it doesn't look practical. I might be talking nonsense. ;-)

michalrus:
@purcell, your suggestions applied! Please, take a look at the current state. A lot cleaner, and works well. =)

The only things to remember in future development are:
* to use `intero-with-temp-buffer` instead of `with-temp-buffer`.
* And, when creating custom non-file buffers, to call `(intero-inherit-local-variables)` at some point.

shlevy:
I had to solve this generally in the support code for nix-buffer https://github.com/NixOS/nixpkgs/blob/282bc88093393e4dcfca68528d4a9d4280a08827/pkgs/build-support/emacs/buffer.nix See specifically https://github.com/NixOS/nixpkgs/blob/282bc88093393e4dcfca68528d4a9d4280a08827/pkgs/build-support/emacs/buffer.nix#L27 where I set advice around `generate-new-buffer` and check for a leading space to identify a temp buffer :open_mouth:

shlevy:
But this approach is much better for a project where you can enforce the use of a custom temp buffer creation function

elisp/intero.el
```diff
@@ -263,6 +263,27 @@ This is slower, but will build required dependencies.")
 (defvar-local intero-buffer-host nil
   "The hostname of the box hosting the intero process for the current buffer.")
 
+(defmacro intero-inherit-local-variables (buffer)
```
shlevy: May want to look at https://github.com/shlevy/inherit-local

elisp/intero.el
```diff
@@ -263,6 +263,27 @@ This is slower, but will build required dependencies.")
...
+  (let ((varlist '(intero-stack-executable
+                   intero-repl-no-build
+                   intero-repl-no-load
+                   ;; TODO: shouldn’t more of the above be here?
+                   )))
+    `(progn
+       ,@(cl-mapcar
+          (lambda (symb)
+            `(setq-local ,symb (buffer-local-value (quote ,symb) ,buffer)))
+          varlist))))
```
purcell: I feel like this doesn't need to be a macro. If you thought it necessary in order to appease `setq-local`, then `(set (make-local-variable 'the-sym) value)` would have the same effect.

michalrus:
@purcell, thank you! Fixed according to your suggestion, looks much simpler, now. :)

purcell:
Great, thanks!

michalrus:
Thank _you_! :bowing_man:

PR #403, opened 2017-04-11
Customize intero-stack-executable-function instead of a variable

michalrus:
Sometimes, the exact path to the executable might not be known, and its name not present in user’s PATH. In such a case, the user would benefit from the possibility of defining their own function to determine that path programmatically.

One possible ‘improvement’ would be to:

```elisp
(defun intero--stack-executable ()
  (funcall intero-stack-executable-function))
```

… and use that everywhere, instead of `(funcal intero-stack-executable-function)`. But I didn’t want to introduce that initially, as another similar symbol might confuse users…

/cc @purcell

Related to #393.

### DIFF
--- a/elisp/intero.el
+++ b/elisp/intero.el
@@ -122,11 +122,16 @@ To use this, use the following mode hook:
   :group 'intero
   :type 'string)
 
-(defcustom intero-stack-executable
-  "stack"
-  "Name or path to the Stack executable to use."
+(defcustom intero-stack-executable-function
+  'intero-stack-executable-default
+  "The function used obtain the name or path of the Stack executable to use."
   :group 'intero
-  :type 'string)
+  :type '(radio (function-item intero-stack-executable-default)
+                (function :tag "Function")))
+
+(defun intero-stack-executable-default ()
+  "Use `stack' by default."
+  "stack")
 
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 ;; Modes
@@ -265,7 +270,7 @@ This is slower, but will build required dependencies.")
 
 (defun intero-inherit-local-variables (buffer)
   "Make the current buffer inherit values of certain local variables from BUFFER."
-  (let ((variables '(intero-stack-executable
+  (let ((variables '(intero-stack-executable-function
                      intero-repl-no-build
                      intero-repl-no-load
                      ;; TODO: shouldn’t more of the above be here?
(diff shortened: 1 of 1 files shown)


michalrus:
This would be very-very-very useful for nixies. :pray:

And it also should be the last change that needs to touch the whole code-base in this shim approach!

purcell:
I'm still not clear on the exact use case for this. When would the shimmed executable be both absent from $PATH and in an unknown location?

Also, when calling `intero-stack-executable-function`, what guarantees would you need to ensure so that implementations of that function could behave sensibly? ie. would you need to ensure `default-directory` is in the Haskell project, rather than in the temp directory etc.?

All in all, I don't really have a clear picture of the problem you're attempting to solve here.

michalrus:
Sure, that’s completely understandable! Sorry for being a bit… enigmatic. :sweat_smile: 

The exact use case I have in mind is this:

* the user doesn’t have their Nix shim for Stack globally in PATH (that should be quite common on NixOS, or really just with Nix-style development),
* and either:
    * they have their project configured so that the executable will be in PATH once they jump into `nix-shell` for their project,
    * or they want the shim to be automatically installed by Nix, once they open any of the project files.

I imagine that last case to be the most common. Common enough, that I shyly thought of adding another (ca. 4-line) alternative implementation for `intero-stack-executable-function` to `intero.el` that would do just that.

As for guarantees, in that last case, this function could be called anywhere, really, as it could just do `nix-shell -p haskellPackages.intero-nix-shim --run 'command -v intero-nix-shim'` or something equivalent<sup>(*)</sup>.

I have almost all the pieces (including proper shim in Haskell), what’s left is just putting them all together and publishing a simple README, of course! (And I figured, I could start with this last change to `intero.el`.)

<sup>(*)</sup> If you decided to reject this PR, the user could be instructed to run this piece of code in `.dir-locals.el`, but that would be significantly more complicated and error-prone than just setting `intero-stack-executable-function` to our correct `intero-stack-executable-nix-shim` or similar.

purcell:
A less intrusive alternative might be to publish a separate package, e.g. `intero-nix`, which would provide a minor `intero-nix-mode`. The user would be told to add `'intero-nix-mode` to their `intero-mode-hook`, so that it would be triggered in any intero-enabled Haskell buffer. `intero-nix-mode` would perform any necessary magic and set `intero-stack-executable` accordingly.

In other words, ideally we'd find a way to use standard elisp patterns to extend `intero` in this way.

michalrus:
Uh, you’re right, from Intero perspective alone, I definitely prefer a simple string. This PR over-complicates the matter… a bit.

OK, I’ll try to come up with something different, potentially following your suggestion in the end. =)

PR #404, opened 2017-04-13
Implement find-uses (display and rename)

chrisdone:
This functionality is borrowed from  the code I wrote for haskell-mode.

This works great for me, super useful for refactoring names and highlighting uses. As-implemented has worked for me for a long time. But the keybindings might conflict with people's expectations.

screenflow

@purcell Do you recommend some more appropriate keybindings?

I once tried integrating this with multiple-cursor-mode but generally I have trouble with that minor mode (sadly).

### DIFF
--- a/elisp/intero.el
+++ b/elisp/intero.el
@@ -179,6 +179,7 @@ To use this, use the following mode hook:
 
 
 (define-key intero-mode-map (kbd "C-c C-t") 'intero-type-at)
+(define-key intero-mode-map (kbd "M-?") 'intero-uses-at)
 (define-key intero-mode-map (kbd "C-c C-i") 'intero-info)
 (define-key intero-mode-map (kbd "M-.") 'intero-goto-definition)
 (define-key intero-mode-map (kbd "C-c C-l") 'intero-repl-load)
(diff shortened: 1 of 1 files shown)


purcell:
This looks really cool. In general, this functionality ties closely into `xref`, which is new in Emacs 25.1: `xref` unifies all the old ad-hoc tag jumping behaviour and jump histories, but it also adds `xref-find-references`, and has query-replace in results.

So two things occur to me here: firstly, in a (possibly not-too-distant) future where we're happy to require Emacs 25.1, we might be able to use the `xref` machinery, and simply provide an `intero` xref backend. (Note that `xref` normally displays found references in a separate window, and query/replace also happens there: I can understand why you've gone for the overlay approach, but piggybacking on Emacs standards would arguably be a good move, because then people can write separate packages which provide a different UX for `xref` globally.)

And secondly, in the short term we might like to use `M-?` as the binding for `intero`'s find-references, since that's what `xref` uses.

chrisdone:
That's cool.  I think we could provide that as a progressive enhancement!

chrisdone:
I've improved the code somewhat.

I've also changed the bindings to copy xref's choice of keybindings. That will naturally flow into xref when we decide to require newer Emacs.

PR #416, opened 2017-05-04
Changes to work with GHC 8.2.1-rc1 API

AaronFriel:
Todo:

- [x] Compiles on GHC 8.2.1-rc2 
- [x] Compiles on GHC 8.0.2
- [x] Compiles on GHC 7.10.3

### DIFF
--- a/cbits/hschooks.c
+++ b/cbits/hschooks.c
@@ -34,7 +34,7 @@ defaultsHook (void)
     // This helps particularly with large compiles, but didn't work
     // very well with earlier GHCs because it caused large amounts of
     // fragmentation.  See rts/sm/BlockAlloc.c:allocLargeChunk().
-    RtsFlags.GcFlags.heapSizeSuggestionAuto = rtsTrue;
+    // RtsFlags.GcFlags.heapSizeSuggestionAuto = rtsTrue;
 #else
     RtsFlags.GcFlags.heapSizeSuggestion = 6*1024*1024 / BLOCK_SIZE;
 #endif
--- a/cbits/hschooks.c
+++ b/cbits/hschooks.c
@@ -30,11 +30,11 @@ initGCStatistics(void)
 void
 defaultsHook (void)
 {
-#if __GLASGOW_HASKELL__ >= 707
+#if __GLASGOW_HASKELL__ >= 707 && __GLASGOW_HASKELL__ < 802
     // This helps particularly with large compiles, but didn't work
     // very well with earlier GHCs because it caused large amounts of
     // fragmentation.  See rts/sm/BlockAlloc.c:allocLargeChunk().
-    // RtsFlags.GcFlags.heapSizeSuggestionAuto = rtsTrue;
+    RtsFlags.GcFlags.heapSizeSuggestionAuto = rtsTrue;
 #else
     RtsFlags.GcFlags.heapSizeSuggestion = 6*1024*1024 / BLOCK_SIZE;
 #endif
--- a/src/GhciFind.hs
+++ b/src/GhciFind.hs
@@ -50,7 +50,11 @@ findQualifiedSource importDecls sample =
             Just name
           | otherwise = Nothing
           where name = unLoc (ideclName m)
+#if __GLASGOW_HASKELL__ >= 802
                 asName = fmap (moduleNameString . unLoc) . ideclAs
+#else
+                asName = fmap moduleNameString . ideclAs
+#endif
 
 -- | Find completions for the sample, context given by the location.
 findCompletions :: (GhcMonad m)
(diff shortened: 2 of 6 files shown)


AaronFriel:
@purcell Uhm, hi, I'm just going to tag you because you recently committed, and I don't know the contributor standards/contributor license agreement stuff on this project. I'd like to know what (if anything) I should do at this point. This is a very "Works On My Machine" patch, and I'm not at all familiar with the Intero internals. I know that I don't know enough to make this a much better patch.

purcell:
I've been sticking mostly to the elisp side of things, and I guess one of the haskell-side committers would need to pick this up and review/sponsor the change, sorry.

cbits/hschooks.c
```diff
@@ -34,7 +34,7 @@ defaultsHook (void)
     // This helps particularly with large compiles, but didn't work
     // very well with earlier GHCs because it caused large amounts of
     // fragmentation.  See rts/sm/BlockAlloc.c:allocLargeChunk().
-    RtsFlags.GcFlags.heapSizeSuggestionAuto = rtsTrue;
```
chrisdone: I think this ought to have a CPP flag for backwards-compat.

src/GhciFind.hs
```diff
@@ -50,7 +50,7 @@ findQualifiedSource importDecls sample =
             Just name
           | otherwise = Nothing
           where name = unLoc (ideclName m)
-                asName = fmap moduleNameString . ideclAs
+                asName = fmap (moduleNameString . unLoc) . ideclAs
```
chrisdone: This should have a CPP if.

src/GhciFind.hs
```diff
@@ -364,7 +364,7 @@ findType infos fp string sl sc el ec =
                             Nothing -> return (FindType minfo ty)
                   _ ->
                     fmap (FindType minfo)
-                         (exprType string)
+                         (exprType TM_Inst string)
```
chrisdone: Again, needs CPP.

src/GhciMonad.hs
```diff
@@ -321,14 +321,14 @@ runDecls decls = do
...
-resume :: (SrcSpan -> Bool) -> GHC.SingleStep -> GHCi GHC.RunResult
+resume :: (SrcSpan -> Bool) -> GHC.SingleStep -> GHCi GHC.ExecResult
 resume canLogSpan step = do
   st <- getGHCiState
   reifyGHCi $ \x ->
     withProgName (progname st) $
     withArgs (args st) $
       reflectGHCi x $ do
-        GHC.resume canLogSpan step
+        GHC.resumeExec canLogSpan step
```
chrisdone: This whole function should also have some CPP.

AaronFriel:
@chrisdone:
> The way to do this is to write e.g. ...
> 
> ```
> #if __GLASGOW_HASKELL__ >= 802
> new code
> #else
> old code
> #endif
> ```

This would need to be `821`, correct, due to the changes in the GHC API? Anyhow, it did not work for me. Perhaps it was just the version of GHC 8.2.1-rc1 I was using? I tried using ">= 821" on another project.

cbits/hschooks.c
AaronFriel: Is this flag no longer used in 8.2.1? I could find almost no documentation on it.

chrisdone:
@AaronFriel I believe the GHC number 802 refers to 8.2 and that you cannot test for the minor version because it _should_ not matter. See here:

https://downloads.haskell.org/~ghc/7.0.3/docs/html/users_guide/version-numbering.html

> The value of __GLASGOW_HASKELL__ (see Section 4.11.3, “Options affecting the C pre-processor”) for a major release x.y.z is the integer xyy (if y is a single digit, then a leading zero is added, so for example in version 6.8.2 of GHC we would have __GLASGOW_HASKELL__==608).

vagarenko:
Can't compile it with GHC-8.2.0rc2
```
src\InteractiveUI.hs:3461:37: error:
    * Variable not in scope:
        utf8DecodeString :: Ptr b0 -> Int -> IO String
    * Perhaps you meant one of these:
        `utf8EncodeString' (imported from Encoding),
        `utf8DecodeByteString' (imported from Encoding),
        `utf8DecodeStringLazy' (imported from Encoding)
     |
3461 |                         $ \(p,n) -> utf8DecodeString (castPtr p) n
     |                                     ^^^^^^^^^^^^^^^^
```
Seems like `utf8DecodeString` was removed from GHC API.

vagarenko:
Yes, it was removed here:
https://github.com/ghc/ghc/commit/1cc82d38759c7a5f527ccc6cb514b8ba576cc3d1

AaronFriel:
@chrisdone Oh well that makes sense!

AaronFriel:
I don't understand how you have "-Werror" in that test script, I can't run it without it erroring on minor warnings in dependencies, e.g.: tabs in `random`.

chrisdone:
I'm going to try adding 8.2.1 to travis.

chrisdone:
Great work @AaronFriel. People will be very happy when 8.2.1 lands. 👏 😄

purcell:
Yes, thanks Aaron!

AaronFriel:
Thank you!

PR #421, opened 2017-05-13
Touch before load

### LINKED ISSUE #143: Intero suppresses GHC warnings from flycheck.
With only haskell-mode and flycheck, I can see lots of warnings on my haskell source codes.
If I add intero, all warnings are suppressed.

How can I see warnings with intero?

chrisdone:
Fixes #143.

I copied the code from turtle. This function is present in System.Directory but only in a newer version than the range we support.

I've had successful builds on OS X, Linux and Windows for this code so it seems portable.

The code is very simple. Just touch each argument passed to `:load` before doing the load procedure. Nothing fancy.

### DIFF
--- a/intero.cabal
+++ b/intero.cabal
@@ -63,6 +63,7 @@ executable intero
     GhciTypes
     GhciInfo
     GhciFind
+    GhciIO
   build-depends:
     base < 5,
     array,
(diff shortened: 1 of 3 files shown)


chrisdone:
It seems that the resolution of utime is in seconds, not milliseconds. So this function isn't sufficient to break the race condition.

chrisdone:
@purcell I think we'll have to do it on the Emacs side.

PR #422, opened 2017-05-13
Cmd queue

chrisdone:
Sadly, like https://github.com/commercialhaskell/intero/pull/421, this doesn't work either.

Test: make a buffer with a warning. Tab enter twice at the end of the buffer. No warnings displayed.

Even though it writes the file just before running the `:l` and each `:l` is sent in isolation rather than all at once.

So it seems that in both cases, the modification time (I suppose at this resolution) is not enough for GHCi/intero to actually do a recompilation.

```
[Intero] -> :l /Users/chris/Emacs/packages/intero/.stack-work/intero/intero377CNO.hs
[Intero] <- [1 of 1] Compiling Demo             ( /Users/chris/Emacs/packages/intero/.stack-work/intero/intero377CNO.hs, /Users/chris/Emacs/packages/intero/.stack-work/intero/intero377p5I/Demo.o )

/Users/chris/Emacs/packages/intero/.stack-work/intero/intero377CNO.hs:5:1: Warning:
[Intero] <- 
    Top-level binding with no type signature: x :: Integer

/Users/chris/Emacs/packages/intero/.stack-work/intero/intero377CNO.hs:5:7: Warning:
    Defaulting the following constraint(s) to type ‘Integer’
      (Num a0) arising from a use of ‘*’
    In the expression:
[Intero] <-  1 * 2
    In an equation for ‘x’: x = 1 * 2
Ok, modules loaded: Demo.
Collecting type info for 1 module(s) ... 

/Users/chris/Emacs/packages/intero/.stack-work/intero/intero377CNO.hs:5:1: Warning:
    Top-level binding with no type signature: x :: Inte
[Intero] <- ger

/Users/chris/Emacs/packages/intero/.stack-work/intero/intero377CNO.hs:5:7: Warning:
    Defaulting the following constraint(s) to type ‘Integer’
      (Num a0) arising from a use of ‘*’
    In the expression: 1 * 2
    In an equation for ‘x’
[Intero] <- : x = 1 * 2

[Intero] -> :l /Users/chris/Emacs/packages/intero/.stack-work/intero/intero377CNO.hs
[Intero] <- Ok, modules loaded: Demo.

[Intero] -> :m + 
[Intero] <- 
```

### DIFF
--- a/elisp/intero.el
+++ b/elisp/intero.el
@@ -225,8 +225,12 @@ and blacklist match, then the whitelist entry wins, and
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 ;; Buffer-local variables/state
 
+(defvar-local intero-current-callback nil
+  "Current callback mutex.")
+
 (defvar-local intero-worker-name nil
   "Name of the current worker.")
+
 (defvar-local intero-callbacks (list)
   "List of callbacks waiting for output.
 LIST is a FIFO.")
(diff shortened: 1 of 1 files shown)


purcell:
> If we're going to have to special-case this, how about we come at it from a different angle and record something like (last-mod-time . flycheck-results) in a buffer-local, and use it to replay the previous flycheck results if we detect the last-mod-time hasn't changed?

I tried implementing this approach locally, only to be stymied by the fact that the mod-times _are_ constantly getting updated, as per my [comment on the original issue](https://github.com/commercialhaskell/intero/issues/143#issuecomment-301298600), so caching flycheck results based on mod-time is not a solution in itself.

PR #423, opened 2017-05-14
Cache flycheck results when the source is unchanged

purcell:
This addresses #143, wherein flycheck warnings were getting lost after two flycheck passes in quick succession since the second underlying :l command simply said "ok".

The approach taken here is to note the temp buffer modification time and results for a flycheck pass, and then to simply replay them if another pass is requested with the same temp buffer mod time.

In order to make this work, we have to ensure that the mod time of the temp buffer is not updated unless its contents change, which introduces a small inefficiency: the temp file is now first quickly read before conditionally writing to it.

### DIFF
--- a/elisp/intero.el
+++ b/elisp/intero.el
@@ -555,6 +555,20 @@ running context across :load/:reloads in Intero."
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 ;; Flycheck integration
 
+(defvar-local intero-check-last-mod-time nil
+  "Most recent modification time of the current buffer when flycheck was last triggered.")
+
+(defvar-local intero-check-last-results nil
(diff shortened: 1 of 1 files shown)


chrisdone:
Tested it. Works great!

purcell:
Yay, great.

fosskers:
This has the unintended side-effect of keeping borked flycheck results when you add a new target via `intero-targets` (say, a test suite) after having already opened the file once. `intero-restart` used to make the false flycheck results go away, but now they stay until you change something.

chrisdone:
Good point. We could perhaps invalidate our caches when we do an intero restart (which is our usual "invalidation point" for other things too).

purcell:
Yep, maybe note the start time for the backend, and ignore cached results if they're older? Saves proactively iterating through all associated buffers after a restart and changing stuff.

purcell:
Fixed in 065f2e7, I believe. Ping me if you find any issues.

fosskers:
Thanks guys.

PR #434, opened 2017-06-18
Added Clang to nix packages.

seanparsons:
Appears this was missing as I couldn't compile the project without it.

Connected to this (and I don't actually know the answer), should stack.yaml be part of the tarball? I'm wondering because as a NixOS user when intero attempts to build itself it does it without honouring any of the nix options and then fails because there's no curses library available.

### DIFF
--- a/stack.yaml
+++ b/stack.yaml
@@ -5,4 +5,4 @@ extra-deps: []
 flags: {}
 extra-package-dbs: []
 nix:
-  packages: [ ncurses ]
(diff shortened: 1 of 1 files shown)


seanparsons:
Is there any reason why this shouldn't be added/

purcell:
> Is there any reason why this shouldn't be added?

I'm not sure, personally, but I'm happy to go ahead and merge this. Hopefully one of the other contributors can answer your `stack.yaml` question.

PR #438, opened 2017-07-03
eldoc-documentation-function: change setq-local to add-function/remove-function

MaskRay:
This makes it composable with other `eldoc-documentation-function` providers.

This snippet is used in other plugins:

https://github.com/leoliu/ggtags/blob/master/ggtags.el#L2246
https://github.com/abingham/emacs-ycmd/blob/master/ycmd-eldoc.el#L222


purcell:
Thanks

Issue #298, opened 2016-10-26
ability to set stack.yaml


zkessin:
It would be very helpful if there was a way to setup intero to use a non default stack.yaml file, esp on projects that may have more than 1 stack.yaml or the like

AshtonKem:
Looks like this is already being looked at in #105.

jml:
#440 fixed this. It can be closed.

Issue #211, opened 2016-08-02
Intero falls over for silly project structure.


angerman:
I have two packages I'm working on, and have them as subfolders of a project that acts as the playground, exercise area for both.

E.g.

```
Playground
|-- PkgA
'-- PkgB
```

Playground simply includes `src`, `PkgA/src` and `PkgB/src`. However both packages sport their own `stack.yaml`. Editing a file in either Pkg, results in intero trying to start a session for that package. In this case PkgB depends on PkgA and intero topples over with ghci trying, but failing to locate PkgA, when editing a file in PkgB.

chrisdone:
If they use eachother, why do you have two separate stack.yaml files?

mgsloan:
I've also encountered this situation.  One case where it occurs is if you have a submodule which contains a stack.yaml.  The resolution doesn't seem clear, though.  One option is to instead default to the outermost `stack.yaml`, but this is inconsistent with the normal CLI behavior.  Perhaps the circumstance where there are multiple `stack.yaml` files in parent directories should be pointed out to the user, and they can select?

angerman:
As @mgsloan it also happens if you have submodules that happen to have `stack.yaml` files. 
My current solution is to look for all `**/stack.yaml` and rename them to `stack.yaml.bck`.
That is a bit annoying but works for now.

chrisdone:
We could provide a `M-x intero-root` or `M-x intero-stack-yaml` command which presents a choice of the detected `.yaml` files. When running any stack command, it will pass `--stack-yaml` (or w/e it was) to the executable, or set it as an environment variable in local buffers. Then some message could say "Use `M-x intero-root` to pick which one."

angerman:
If we can inhibit intero from trying to start a second session, that would solve all my issues.

E.g. "You already have an intero session running for a project higher up in the hierarchy, however, the file you are editing belongs to project with a separate `stack.yaml`. Do you want to start a separate intero session? Y/N"

angerman:
I guess the simples would be: *if* an intero session is already running in a higher up in the fs hierarchy, do *not* start a new intero session. If one want's a separate session `C-u M-x intero-repl`?

ntc2:
I would like something like the `intero-root` that @chrisdone suggests, or even a `intero-stack-yaml-blacklist`.

[...]

PR #440, opened 2017-07-03
Add ability to specify stack yaml file to use (closes #105, #298)

### LINKED ISSUE #105: In elisp, allow a specification of which stack.yaml to use
It' be convenient to have a way to switch which stack config you're using (the `--stack-yaml` flag).  This'd be particularly convenient if it's provided by filepath selection filtered to `.yaml`, starting in the dir of the current stack config.

Since targets are provided directly to stack, it turns out that you can already pass `--stack-yaml stack-7.8.yaml` in via `intero-targets`.  However, this does not work with intero's auto-install, because it doesn't know to target GHC 7.8, and instead we can get a compiler mismatch if the primary `stack.yaml` uses a different ghc.

timds:
Hi, first-time contributor here!

This commit adds the ability to switch yaml config files using `M-x intero-stack-yaml`. Intero still works as before by assuming the default stack.yaml / global file, but now you can change it afterwards. You can optionally preserve that change across sessions. Switching yaml files restarts intero and intero-repl. 

I would appreciate if anyone can look at my this commit and let me know if it's OK. 

I have not addressed #211, however I think this commit is a step towards that.

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,7 +28,11 @@ To enable `intero` in all `haskell-mode` buffers by default, enable
 `(intero-global-mode 1)` to your Emacs start-up files.
 
 Intero will then activate for all projects, and for files without a
-stack.yaml, it will assume the "global" project.
+stack.yaml, it will assume the "global" project. If you want to use an
+alternate stack yaml configuration file (for example, when developing
+for multiple GHC versions), use `M-x intero-stack-yaml` to switch
+file. When switching configuration, you will asked whether you want to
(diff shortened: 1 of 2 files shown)


purcell:
This is awesome, and looks tidy and well-considered. If it weren't a big change, I'd go ahead and merge it right away. @chrisdone - you okay with this?

purcell:
_Presses "merge" button and runs away_

Issue #441, opened 2017-07-06
":complete repl" missing feature


KarlFish:
In `stack ghci` running in a terminal I get this interaction:

    ghci> let vvvvvvvvvvvvvvvvvvvvvvvvvvvv = 10
    ghci> :complete repl "vvvvv"
    1 1 ""
    "vvvvvvvvvvvvvvvvvvvvvvvvvvvv"
    ghci> 

but in intero repl mode I get this:

    λ let vvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvv = 10
    λ :complete repl "vvvvv"
    0 0 ""
    λ

[This issue](https://github.com/commercialhaskell/intero/issues/245) mentioned that the autocomplete didn't work, and the fix used `:complete` in `intero-completion-at-point`. However `:complete` doesn't work correctly in intero repl. 

The source for `completeCmd :: String -> GHCi ()` is the same across ghci and intero.

KarlFish:
I think it might have something to do with setting ghciState at [https://github.com/commercialhaskell/intero/blob/adb30bfcd4b280005dc6bf505cb9ff0326f7fbc5/src/InteractiveUI.hs#L655](https://github.com/commercialhaskell/intero/blob/adb30bfcd4b280005dc6bf505cb9ff0326f7fbc5/src/InteractiveUI.hs#L655)

KarlFish:
I'm working on figuring out why, but ghc takes forever to compile.

KarlFish:
The ghci I built from source has user defined expressions appended to the list returned by `GHC.getRdrNamesInScope`, while the one built from intero does not.

KarlFish:
The issue is in `completeIdentifier`

https://github.com/commercialhaskell/intero/blob/master/src/InteractiveUI.hs#L2941

I will create a pull request.

PR #442, opened 2017-07-06
Improve auto complete at point

KarlFish:
Fixes https://github.com/commercialhaskell/intero/issues/441 in the local `stack build && exec intero` repl. I couldn't figure out how to test my local build as an intero-global-mode backend with emacs. 

With this removed, there are no uses of `rdrNamesInScope` in the entire project. 

    ~/programming/intero $ ag -a rdrNamesInScope
    src/InteractiveUI.hs
    564:                   rdrNamesInScope     = [],
    655:  modifyGHCiState (\s -> s { rdrNamesInScope = names })
    1570:      lift (modifyGHCiState (\s -> s { mod_infos = newInfos, rdrNamesInScope = names }))
    2307:  modifyGHCiState (\s -> s { rdrNamesInScope = names })

    src/GhciMonad.hs
    126:        rdrNamesInScope :: ![GHC.RdrName],

I think this field was meant to allow autocompletion in the main buffer of the `intero-global-mode` even when the modules don't compile (but did at one point). I would have tested how this pull request changes that, but I couldn't figure out how to.

PR #444, opened 2017-07-07
Fix REPL autocomplete

KarlFish:
A redo of my effort to fix https://github.com/commercialhaskell/intero/issues/441 ([original](https://github.com/commercialhaskell/intero/pull/442)).

There are two lists of scopes available in the code, `GHCiState` and `GHC.getRdrNamesInScope`. They start off the same. `GHCiState` does not change when the user adds new identifiers into the scope through let bindings in the repl, but if the user `:load`s a file with errors in it, `GHC.getRdrNamesInScope` is cleared. 

~~My proposed solution is to lazily combine the pros of each. I think the idea here is on target, but they style isn't very good yet.~~ Edit: solution changed with `git push --force`

### DIFF
--- a/example.hs
+++ b/example.hs
@@ -1,0 +1,1 @@
+fail = 3 :: String
--- a/example.hs
+++ b/example.hs
@@ -1,1 +1,0 @@
-fail = 3 :: String
--- a/src/GhciMonad.hs
+++ b/src/GhciMonad.hs
@@ -123,7 +123,6 @@ data GHCiState = GHCiState
 
         -- stored state
         mod_infos :: !(Map ModuleName ModInfo),
-        rdrNamesInScope :: ![GHC.RdrName],
 
         ghci_work_directory :: FilePath,
             -- ^ Used to store the working directory associated with
--- a/src/InteractiveUI.hs
+++ b/src/InteractiveUI.hs
@@ -592,7 +592,6 @@ interactiveUI config srcs maybe_exprs = do
                    short_help          = shortHelpText config,
                    long_help           = fullHelpText config,
                    mod_infos           = M.empty,
-                   rdrNamesInScope     = [],
                    ghci_work_directory = current_directory,
                    ghc_work_directory  = current_directory
                  }
@@ -680,10 +679,6 @@ runGHCi paths maybe_exprs = do
 
   -- reset line number
   getGHCiState >>= \st -> setGHCiState st{line_number=1}
-
-  -- Get the names in scope
-  names <- GHC.getRdrNamesInScope
-  modifyGHCiState (\s -> s { rdrNamesInScope = names })
 
   case maybe_exprs of
         Nothing ->
@@ -1635,11 +1630,10 @@ doLoad retain_context howmuch = do
       return ok
   case wasok of
     Succeeded -> do
-      names <- GHC.getRdrNamesInScope
       loaded <- getModuleGraph >>= filterM GHC.isLoaded . map GHC.ms_mod_name
       v <- lift (fmap mod_infos getGHCiState)
       !newInfos <- collectInfo v loaded
-      lift (modifyGHCiState (\s -> s { mod_infos = newInfos, rdrNamesInScope = names }))
+      lift (modifyGHCiState (\s -> s { mod_infos = newInfos }))
     _ -> return ()
   return wasok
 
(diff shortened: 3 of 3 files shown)


KarlFish:
I opened https://github.com/commercialhaskell/intero/issues/441 because the intero repl in emacs wouldn't complete any variables I added into scope. 

This pull request makes auto-complete in intero work the same way as in `stack ghci`. 

My core change - https://github.com/commercialhaskell/intero/pull/444/files#diff-a469118375854c9ae9408cbd3bb03b39R2932

GHCi source - https://github.com/ghc/ghc/blob/master/ghc/GHCi/UI.hs#L3028

KarlFish:
I just got emacs to use the intero I built from source, and I found an issue with my pull request. Here's an example.

    -- In the modified intero repl, in emacs
    λ let sinxxxxxxxxxxxxxxxx = 10
    λ sin<cursor>

`M-x completion-at-point` (tab autocomplete) lists `sin,sinh`
`M-x hippie-expand` inserts sinxxxxxxxxxxxxxxxx. Repeated calls cycle through the other options.
Typing `:complete repl "sin"\n` returns `sin, sinh, sinxxxxxxxxxxxxxxxx`. 

This pull request does not fix much yet.

KarlFish:
`hippie-expand ` expands any words already typed, which doesn't help for ghci-like autocomplete

The intero completion-at-point function uses async calls to a separate repl process, one that doesn't have the variables introduced, so the visible repl isn't cluttered with `:complete`'s.

Evaluating `(comint-simple-send (intero-process 'backend) "let sinyyyyyyyyyyyyy = 10")` while in the intero repl buffer, makes `completion-at-point` willing to expand `sin` to `sinyyyyyyyyyyyyy`.

KarlFish:
This is a deliberate design decision, so this information is already known. I'm writing these comments for anyone else who was wondering why intero autocomplete doesn't work right. 

I'll keep tinkering around to see if anything comes up. I'm willing to close this issue if I give up for a while.

KarlFish:
The ideal would be to send `:complete` commands to the main process buffer, but intercept the output so it doesn't clutter up the repl. `comint-preoutput-filter-functions` looks like the key here. 

    Documentation:
    List of functions to call before inserting Comint output into the buffer.
    Each function gets one argument, a string containing the text received
    from the subprocess.  It should return the string to insert, perhaps
    the same string that was received, or perhaps a modified or transformed
    string.

KarlFish:
A quick test proves promising.

### Changes
Insert the following after https://github.com/commercialhaskell/intero/blob/master/elisp/intero.el#L1279 

       (add-hook 'comint-preoutput-filter-functions
                'my-func
                t t)
 
Where my-func is defined as:

    (defun my-func (string-thing)
      (message (format "Output was: %S" string-thing))
      "And now for something completely different!")



### Results


After running `intero-repl-load` (C-c C-l), a few `And now for something completely different!`s appear. 

Typing `3<return>` adds two more to the repl output. The message buffer shows `my-func` received `"3\n"` and `^D "`. 

Typing `:complete repl sin<return>` adds one more copy to the buffer. The message buffer shows `my-func` received:

    "2 2 \"\"
    \"sin\"
    \"sinh\"
    ^D "

`^D` is the lambda symbol (the repl prompt). Typing `C-q C-d` into the repl buffer outputs a lambda.

### Implications

A better `my-func` would intercept output in the exact format of `:complete`. `intero-completion-at-point` may have to use a different async call (or use a worker called `'frontend) to make sure it uses the main repl buffer comint process.

I will continue to look into this.

KarlFish:
### Update

`comint-preoutput-filter-functions` is annoying because if the input is too long, it processes it with several calls to the functions. This lazyness means we need to keep a record of the previous ouputs once the ouput starts matching the `:complete repl` format. 

I am working on an alternate idea of sending `:complete`s to the current repl's process (instead of `'backend`'s by changing `intero-completion-at-point`, and redirecting the output to a temporary buffer with `comint-redirect-send-commmand`.

Ideally I could leave `intero-completion-at-point` alone, and just add another branch to [`intero-completions-grab-prefix`](https://github.com/commercialhaskell/intero/blob/master/elisp/intero.el#L890).

KarlFish:
I'm realizing just now that my commentary on this process should be in the issue I opened instead of a pull request. I started updated this page when I realized the pull request didn't actually fix anything (it still doesn't), but I didn't want to close it and open a third when I thought I was ready again. 

I'm going to continue updating this request (unless told otherwise) to keep all the information I find in one place. My next pull request will be neater.

KarlFish:
I now have a version of `intero-completion-at-point` that uses native GHCi tab completion in the REPL. Typing `C-M-i` in intero functions (as far as I can tell) identically to typing `<tab>` in ghci.

[gist of `intero-completion-at-point`](https://gist.github.com/KarlFish/1845364b863a030c01487d340ed33310)

KarlFish:
This is a mess. Why does it repeat the first two commits, add two more, and then merge the branch back together? 

I'm closing this pull request and open a much cleaner one.

Issue #245, opened 2016-08-28
Autocompletion not working in REPL


mtesseract:
Hi,

unless I am overlooking something obvious, there is no tab-autocompletion in the Intero REPL.
It would be very nice to have, though. :-)

Thanks,
Moritz

chrisdone:
I recently pushed support so that company-mode is activated in the REPL. Have you tried a recent version?

crocket:
I installed the latest version from melpa today, but tab autocompletion doesn't work.
When I press TAB, a lot of spaces are inserted into REPL.

mtesseract:
Sorry for the delay.

The melpa version intero-20160825.656 didn't provide me with tab-completion in the REPL (it simply inserts whitespaces).

I have just directed my Emacs (24.5.1) to use intero's git master version, it has shown the same behaviour.
C-h k TAB tells me:

TAB (translated from <tab>) runs the command indent-for-tab-command,
which is an interactive compiled Lisp function in `indent.el'.

[I have the following snippet in my .emacs for using intero's master:
(add-to-list 'load-path "~/.emacs.d/intero/elisp/") ;; This is the master branch
(require 'intero)
(add-hook 'haskell-mode-hook 'intero-mode)
]

Am I doing anything wrong here? Can I help debugging this?

ntc2:
Same problem: `company-mode` completions do work in the REPL, but **plain old TAB completions**, like you get with terminal GHCi, **do not work**.

Running `C-h k TAB` in the REPL tells me TAB is bound to `indent-for-tab-command`. I tried inserting a raw TAB with `C-q TAB`, but that doesn't work either ?!

TimoFreiberg:
I assume I have the same problem: I can use company mode for completions, but the completions seem to be pure text-based. I don't get suggestions based on imported modules, for example.

The old haskell-interactive mode shows completion suggestions from all imported packages, the entire Prelude, etc.
I miss this in Intero

pashosh:
i miss it also :-)

purcell:
@TimoFreiberg Not the same issue, so it might be a separate bug, or a point of confusion. You'll only see company completions for imported modules in the Intero REPL when you've successfully loaded a Haskell file.

@pashosh, @ntc2, @crocket and @mtesseract:

[...]

PR #454, opened 2017-08-01
Fix repl autocomplete

KarlFish:
I was frustrated that intero auto-complete suggestions did not update as I added new identifiers. The solution given at issue https://github.com/commercialhaskell/intero/issues/245 sends `:complete repl` calls to an untouched back-end instance of intero, instead of the visible instance that has all of the identifiers defined.

I changed `intero-completion-at-point` to send commands to the visible instance so tab completion works like the native version.

I also fixed a related issue in the intero repl backend where running `:complete repl "<"` lists every single identifier in scope instead of just those that start with `<` in https://github.com/commercialhaskell/intero/commit/68b41c66596bc98bd533c4c1ed6f2602bcef3399

I would like constructive feedback on `intero-completion-at-point` (now named `intero-repl-tab-completion-at-point`). I think my ad-hoc async call should be integrated with `intero-async-call`. 

I assume people working in the repl would not need to use tab to indent so in https://github.com/commercialhaskell/intero/commit/24dc2be401c823102f069e40aeb89bc75f560d09 I bound tab directly to `'complete-symbol` in the repl mode. Is that appropriate?

---------------------------------------------
Steps to run emacs with a locally installed intero repl: 
1. Compile intero from source
2. Copy `intero/elisp/intero.el` to `~/.emacs.d/elpa/intero-<number>/intero.el` (or its equivalent)
3. Make a new project, set the resolver to lts-5.6
4. `stack setup` and `stack build` new project
5. Open project source file in emacs with intero, load repl, and exit (if you get a warning that mentions ignored, `chmod 700 testproject/.stackwork/intero`)
6. Copy intero binary to replace `testproject/.stackwork/install/<platform>/lts-5.6/7.10.3/bin/intero` (or its equivalent)

### DIFF
--- a/elisp/intero.el
+++ b/elisp/intero.el
@@ -807,25 +807,21 @@ CHECKER and BUFFER are added to each item parsed from STRING."
     (when prefix-info
       (cl-destructuring-bind
           (beg end prefix _type) prefix-info
-        (let* ((tmp-buf (get-buffer-create (make-temp-name "intero-completions")))
-               (completions
-                (save-excursion
-                  (comint-redirect-send-command
-                   (format ":complete repl %S" prefix) ;; command
-                   tmp-buf ;; output buffer
-                   nil     ;; echo
-                   t)    ;; no-display
-                  (while (not comint-redirect-completed)
-                    (sleep-for 0.01))
-                  (switch-to-buffer tmp-buf)
-                  (let ((str (buffer-string)))
-                    (kill-buffer tmp-buf)
-                    (intero-completion-response-to-list str)))))
-          (let ((first-line (car completions)))
+        (let ((proc (get-buffer-process (current-buffer))))
+          (with-temp-buffer
+            (comint-redirect-send-command-to-process
+             (format ":complete repl %S" prefix) ;; command
+             (current-buffer) ;; output buffer
+             proc ;; target process
+             nil  ;; echo
+             t)   ;; no-display
+          (while (not comint-redirect-completed)
+            (sleep-for 0.01))
+          (let ((first-line (car (buffer-string))))
             (when (and (string-match "[^ ]* [^ ]* " first-line) ;; "2 2 :load src/"
-                       completions)
+                       (buffer-string))
               (setq first-line (replace-match "" nil nil first-line))
-              (list (+ beg (length first-line)) end (cdr completions)))))))))
+              (list (+ beg (length first-line)) end (cdr (buffer-string)))))))))))
 
 (defun intero-text-from-prompt-to-point ()
   "Return the text from this buffer from the prompt up to right behind the point."
--- a/elisp/intero.el
+++ b/elisp/intero.el
@@ -809,7 +809,6 @@ CHECKER and BUFFER are added to each item parsed from STRING."
           (beg end prefix _type) prefix-info
         (let* ((repl-buffer (current-buffer))
                (proc (get-buffer-process (current-buffer)))
-               (time0 (float-time))
                (output
                 (with-temp-buffer
                   (comint-redirect-send-command-to-process
@@ -818,20 +817,14 @@ CHECKER and BUFFER are added to each item parsed from STRING."
                    proc ;; target process
                    nil  ;; echo
                    t)   ;; no-display
-                  (defun done ()
-                    (with-current-buffer repl-buffer
-                      comint-redirect-completed))
-                  (while (and (not (done))
-                              (< (- (float-time) time0) 3))
+                  (while (not (with-current-buffer repl-buffer
+                                      comint-redirect-completed))
                     (sleep-for 0.01))
-                  (if (done)
-                      (let* ((completions (intero-completion-response-to-list (buffer-string)))
-                             (first-line (car completions)))
-                        (when (string-match "[^ ]* [^ ]* " first-line) ;; "2 2 :load src/"
-                          (setq first-line (replace-match "" nil nil first-line))
-                          (list (+ beg (length first-line)) end (cdr completions))))
-                    ; old intero builds interfere with :complete
-                    (message "Comint repl autocomplete timed out")))))
+                  (let* ((completions (intero-completion-response-to-list (buffer-string)))
+                         (first-line (car completions)))
+                    (when (string-match "[^ ]* [^ ]* " first-line) ;; "2 2 :load src/"
+                      (setq first-line (replace-match "" nil nil first-line))
+                      (list (+ beg (length first-line)) end (cdr completions)))))))
           output)))))
 
 (defun intero-text-from-prompt-to-point ()
--- a/elisp/intero.el
+++ b/elisp/intero.el
@@ -808,24 +808,22 @@ CHECKER and BUFFER are added to each item parsed from STRING."
       (cl-destructuring-bind
           (beg end prefix _type) prefix-info
         (let* ((repl-buffer (current-buffer))
-               (proc (get-buffer-process (current-buffer)))
-               (output
-                (with-temp-buffer
-                  (comint-redirect-send-command-to-process
-                   (format ":complete repl %S" prefix) ;; command
-                   (current-buffer) ;; output buffer
-                   proc ;; target process
-                   nil  ;; echo
-                   t)   ;; no-display
-                  (while (not (with-current-buffer repl-buffer
-                                      comint-redirect-completed))
-                    (sleep-for 0.01))
-                  (let* ((completions (intero-completion-response-to-list (buffer-string)))
-                         (first-line (car completions)))
-                    (when (string-match "[^ ]* [^ ]* " first-line) ;; "2 2 :load src/"
-                      (setq first-line (replace-match "" nil nil first-line))
-                      (list (+ beg (length first-line)) end (cdr completions)))))))
-          output)))))
+               (proc (get-buffer-process (current-buffer))))
+          (with-temp-buffer
+            (comint-redirect-send-command-to-process
+             (format ":complete repl %S" prefix) ;; command
+             (current-buffer) ;; output buffer
+             proc ;; target process
+             nil  ;; echo
+             t)   ;; no-display
+            (while (not (with-current-buffer repl-buffer
+                          comint-redirect-completed))
+              (sleep-for 0.01))
+            (let* ((completions (intero-completion-response-to-list (buffer-string)))
+                   (first-line (car completions)))
+              (when (string-match "[^ ]* [^ ]* " first-line) ;; "2 2 :load src/"
+                (setq first-line (replace-match "" nil nil first-line))
+                (list (+ beg (length first-line)) end (cdr completions))))))))))
 
 (defun intero-text-from-prompt-to-point ()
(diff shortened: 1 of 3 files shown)


KarlFish:
For step 3, edit `intero.el` wherever it appears in your `.emacs.d` directory

purcell:
> I assume people working in the repl would not need to use tab to indent so in 24dc2be I bound tab directly to 'complete-symbol in the repl mode. Is that appropriate?

Not really. The goal here should not be to mimic GHCi's keys inside Emacs, but rather to behave consistently with other Emacs `comint`/`repl` modes. So `TAB` handling should be left unchanged.

I'm sorry that I don't have time to review the rest of the change just now, but one quick superficial comment I have is that `-tab-` shouldn't appear in the completion function's name.

KarlFish:
As this project's maintainer you've got to have much higher standards of code than I do. Take as long as you need to make sure it's actually maintainable. It's a lot of changes, and I don't have much elisp experience. 

I tried to keep each commit logically separtate, but that broke down when editing `intero-completion-at-point`.

elisp/intero.el
```diff
@@ -1280,7 +1297,8 @@ function is subsequently applied to each line, once."
             'intero-linkify-process-output
             t t)
   (setq-local comint-prompt-read-only t)
-  (add-hook 'completion-at-point-functions 'intero-completion-at-point nil t)
+  (setq-local company-idle-delay nil)
```
purcell: This line is out of place: we shouldn't be overriding general `company` preferences arbitrarily. What's the intention?

elisp/intero.el
```diff
@@ -802,19 +801,37 @@ CHECKER and BUFFER are added to each item parsed from STRING."
...
-               (intero-completion-response-to-list
-                (intero-blocking-call
-                 'backend
-                 (format ":complete repl %S" prefix)))))
-          (when completions
-            (list beg end completions)))))))
+        (let* ((tmp-buf (get-buffer-create (make-temp-name "intero-completions")))
+               (completions
+                (save-excursion
+                  (comint-redirect-send-command
```
purcell: You were concerned about this, but it looks broadly reasonable to me. :-)

elisp/intero.el
```diff
@@ -151,7 +151,6 @@ To use this, use the following mode hook:
   (if intero-mode
       (progn
         (intero-flycheck-enable)
-        (add-hook 'completion-at-point-functions 'intero-completion-at-point nil t)
```
purcell: This gave me pause, because I guess my intention in adding this line to `intero-mode` was to ensure there was always a sensible `completion-at-point` function in place, even in `.hs` source buffers. But we're pretty much forcing `company` on everyone, so that's fine! And I like that the `completion-at-point` function is now clearly named in order to be specific to the REPL.

elisp/intero.el
```diff
@@ -802,19 +801,37 @@ CHECKER and BUFFER are added to each item parsed from STRING."
...
       (cl-destructuring-bind
           (beg end prefix _type) prefix-info
-        (let ((completions
-               (intero-completion-response-to-list
-                (intero-blocking-call
-                 'backend
-                 (format ":complete repl %S" prefix)))))
-          (when completions
-            (list beg end completions)))))))
+        (let* ((tmp-buf (get-buffer-create (make-temp-name "intero-completions")))
```
purcell: You probably want `with-temp-buffer`, rather than to create a buffer with a temporary name you've generated yourself. (`make-temp-name` is geared towards temporary filenames, anyway, whereas temp buffers typically look like ` *some-temp-thing*`: note the leading space, which excludes it from the buffer list.)

elisp/intero.el
KarlFish: If the repl autocomplete had no completions, I didn't want it to fall through to something else. Reduce inconsistency with the other changes.

elisp/intero.el
KarlFish: It fixed an annoying issue with my config. You're right that it's out of place, it should be in fixed on my end.

elisp/intero.el
KarlFish: Ah. `comint-redirect-send-command` uses the current buffer's process so we need to hold on to a reference to the intero buffer process and call `comint-redirect-send-command-to-process`. My quick fix didn't work so I need think more about nested let bindings.

purcell: Yep, sorry, didn't mention that. :-)

KarlFish:
Starting college distracted me for a bit.

I'm trying to rewrite the `intero-repl-completions-at-point` using `with-temp-buffer`, but I keep getting a weird error, and I can't figure out why.

I've written the function [here](https://gist.github.com/KarlFish/bc381dc85140618fe2bfb9ce35f20245), but  when I test it, (by typing `si<C-M-i>`, I get this error:

```
Debugger entered--Lisp error: (error "Selecting deleted buffer")
  comint-redirect-preoutput-filter("4 \"\"\n\"significand\"\n\"signum\"\n\"sin\"\n\"sinh\"\n ")
  comint-redirect-filter(comint-output-filter #<process intero> "4 \"\"\n\"significand\"\n\"signum\"\n\"sin\"\n\"sinh\"\n ")
  apply(comint-redirect-filter comint-output-filter (#<process intero> "4 \"\"\n\"significand\"\n\"signum\"\n\"sin\"\n\"sinh\"\n "))
  #[128 "\300\301\302#\207" [apply comint-redirect-filter comint-output-filter nil] 5 nil](#<process intero> "4 \"\"\n\"significand\"\n\"signum\"\n\"sin\"\n\"sinh\"\n ")

```

KarlFish:
`intero-reple-completion-at-point` uses `with-temp-buffer`. 

I had to add a reference to the intero repl buffer to figure out when the process completed because I didn't want to loop through `intero-list-buffers` for the right one.

I've tested in a clean emacs setup (only intero installed), and the elisp works whether the intero binary has the changes in this pull request or not.

KarlFish:
@purcell This pull request is marked "Changes requested", but the only changes I haven't updated since then you wrote you were okay with. Let me know if there's anything else that can be improved.

Thank you so much for your patience through this process (especially at the beginning)

purcell:
Thanks @KarlFish: this looks good to me! @chrisdone - can you have a quick scan please and merge the PR if you're also happy?

chrisdone:
@KarlFish Can you confirm my question?

KarlFish:
What question?

chrisdone:
Ah, good job I asked. I was wondering if you saw it. Here:

https://github.com/commercialhaskell/intero/pull/454/files/39bbf2a207b0ee369e8556bbb81187a80b8ffca0#diff-a469118375854c9ae9408cbd3bb03b39R3035

I believe if a load has just failed, then requesting the rdrnames *now* will yield just basic Prelude and nothing from the unsuccessful modules. That was the purpose of caching it via the GHCi state. Can you confirm this isn't a regression? Or perhaps it's not a worry.

I'd like to accept this PR as it's been here a while and looks good, and that seems my only query.

KarlFish:
My thought process was to make the auto-complete more like ghci. I view it as a positive that the repl will not complete to identifies that are no longer in scope. If a load has just failed, ghci dumps all the identifies in scope. Why complete things that don't exist. Bonus: it now completes imports just like ghci. 

This updated function is tied to the state of the repl so I removed it from the generic intero-completion-at-point function hook. Auto-complete in the buffer is much harder.

KarlFish:
I haven't looked into it, but `:type-at` stands out to me as something worth keeping intero for. Autocomplete really stood out to me as a missing features because I like using it to explore what's in scope.

I see replacing intero with Haskell IDE Engine emacs frontend as the longterm goal, intero is kind of a hack that works really well.

KarlFish:
I just like spamming tab and seeing things popup.

chrisdone:
I didn't mean replacing Intero for everything, it has a number of features, bugfixes and backported code that GHCi doesn't have. For Flycheck and buffer completion we have Intero.

For the REPL, however, we just launch a process and don't make use of anything outside of regular GHCi. For the REPL specifically we could just launch GHCi.

KarlFish:
Oh, right. Using ghci instead of intero could mean a user doesn't have to install intero backend if they just want the repl. The auto-complete code won't change too much, but `linkify` hooks could be deleted. I see an overall advantage there

ntc2:
Just wanted to chime in and say that for me, having the Intero REPL behave like GHCi is exactly what I want. Right now I use Intero mode in Emacs for editing, but run `stack ghci` in a separate terminal for my REPL, since the completion and history search in the Intero REPL are different from in GHCi. But being able to use `C-c C-l` to get a REPL with GHCi bindings in Emacs, with my current buffer loaded, would be nice.

From earlier comments from purcell it sounds like TAB won't be bound to completion by default in the Intero REPL even after these changes. So, documentation on how to get GHCi bindings in Intero, or even better an option for GHCi bindings that defaults to off, what be appreciated. Specifically, I use TAB for completion and Ctrl-r for history search a lot in GHCi.

KarlFish:
@ntc2 add `(define-key intero-repl-mode-map (kbd "C-r") 'comint-history-isearch-backward)` to your config. Working on tab completion

chrisdone:
`M-r` does searching backwards already in comint-derived modes!

chrisdone:
OK, then I'll change it to launch `ghci` instead and close this PR.

ntc2:
@KarlFish, @chrisdone : that command, `comint-history-isearch-backward`, doesn't behave quite like `Ctrl-r` in GHCi: hitting `Ctrl-r` multiple times in in GHCi cycles through matches, but doesn't `comint-history-isearch-backward` gets stuck at the first result. Is there some fundamental problem that prevents just passing some keys through to the inferior GHCi process, and letting GHCi handle those keys in the way it does in the shell?

@chrisdone : the point is that I don't want to learn different commands for Intero-GHCi and non-Intero-GHCi. It just seems like extra mental effort with no benefit.

In any case, thank you both for suggesting `M-r` and how to rebind it to `C-r`.

KarlFish:
@chrisdone Changing the repl to use ghci is a much more elegant solution than updating intero binary. But I also made changes to `intero.el` that have not been merged.

`λ import D<tab>` results in `λ import Double` instead of `λ import Data.` because `intero-completion-at-point` uses `intero-completions-grab-prefix` to grab the last token. This is ideal behavior in the main text buffer, but is a breaking change from the ghci repl. I fixed this by adding `intero-text-from-prompt-to-point`.

I also switched `intero.el` to use the repl ghci process instead of the intero  `'backend` process. Currently the following interaction results in `No Match`.
```
λ let veryLongIdentifier = 10
λ ve<tab>
```
I fixed this by changing `intero-completion-at-point` to directly send the string to the ghci process, but it would be more elegant to introduce `'frontend` to compliment `'backend`.

KarlFish:
@ntc2 I don't have the same issue with `comint-history-isearch-backward`

Pressing `C-r` multiple times cycles through the history on my test setup


intero

ntc2:
@KarlFish you're right, my mistake was that I didn't notice the `-regexp` in the `M-r` binding. Using the non-regexp version it works the way I want, like in your screen capture.

PR #459, opened 2017-08-10
Fix tests on Windows

VitorCBSB:
Tests should run correctly on Windows now.

One of the solutions in particular relies on System.Info.os, as it turns out different outputs are generated depending on the OS. Other solutions for that include things like CPP, which I'm not sure are desirable at all.


purcell:
Thanks!

PR #468, opened 2017-09-05
Fix cache invalidation strawman

chrisdone:
@purcell The infamous flycheck disappearing error messages error persisted since we fixed it. At the time I was so busy I just accepted it. It's starting to become a problem for me. I ended up developing a habit of hitting RET just to make the flycheck errors update.

screenflow

Are you not reproducing this? The elisp-side caching we did doesn't work. 

Meanwhile I've come up with an alternative solution:

The problem with the historical approach is that Emacs Lisp writes to the temp file NOW. For the duration of one second, any subsequent updates to the module are ignored by GHCi. And, of course, our code doesn't rewrite the file if the contents didn't change. So there's nothing to invalidate the file afterwards anyway. But even if we did update the file, it'd be ignored for 1 second. So we need a way to sleep between commands without blocking Emacs or having async command queue issues.

Instead, we already have an ordered queue in Intero's query/reply REPL. So we can write the file contents to `STAGING` immediately, and then queue up:

1. We queue up a command for intero to copy it to the temp file when it gets to it in its own time. After copying STAGING to TEMP, that updates TEMP's contents and modified time. We delete STAGING.
2. We queue up :load to load that TEMP file. 
3. Finally, we queue up a third item: sleep for one second. This prevents step (1) from running until after one second has passed.

We retain proper ordering, all calls by flycheck return results pertaining to that exact change, and we don't block the Emacs UI at all with the sleep. We sleep afterwards so that the user doesn't have to wait for the first set of results to come back, just subsequent ones.

**Summary**

I'm going to test this out for a few days, but I have confidence in its dumb simplicity that it's going to solve it. The problem for distributing this would be that I use POSIX commands; so it may rather be better to add these commands to the Intero as `:copy-and-check` and `:sleep`.

### DIFF
--- a/elisp/intero.el
+++ b/elisp/intero.el
@@ -1622,7 +1622,7 @@ path."
     temporary-file-directory))
 
 (defun intero-temp-file-name (&optional buffer)
-  "Return the name of a temp file containing an up-to-date copy of BUFFER's contents."
+  "Return the name of a temp file pertaining to BUFFER."
   (with-current-buffer (or buffer (current-buffer))
     (or intero-temp-file-name
         (progn (setq intero-temp-file-name
--- a/elisp/intero.el
+++ b/elisp/intero.el
@@ -70,7 +70,7 @@
   :group 'haskell)
 
 (defcustom intero-package-version
-  "0.1.21"
+  "0.1.22"
   "Package version to auto-install.
 
 This version does not necessarily have to be the latest version
@@ -661,13 +661,12 @@ running context across :load/:reloads in Intero."
            (staging-file (intero-localize-path (intero-staging-file-name)))
            (temp-file (intero-localize-path (intero-temp-file-name)))
            (hash (intero-check-calculate-hash)))
-      ;; We queue up to mv the staging file to the target temp
-      ;; file and touch it to update its modified time.
+      ;; We queue up to :move the staging file to the target temp
+      ;; file, which also updates its modified time.
       (intero-async-call
        'backend
-       (format ":!mv \"%s\" \"%s\"; touch \"%s\""
+       (format ":move \"%s\" \"%s\""
                staging-file
-               temp-file
                temp-file))
       ;; We load up the target temp file, which has only been updated
       ;; by the copy above.
@@ -705,7 +704,7 @@ running context across :load/:reloads in Intero."
       ;; changed, if the timestamp is less than 1 second.
       (intero-async-call
        'backend
-       ":!sleep 1"))))
+       ":sleep 1"))))
 
 (flycheck-define-generic-checker 'intero
   "A syntax and type checker for Haskell using an Intero worker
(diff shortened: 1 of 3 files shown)


chrisdone:
This could be optimized as a `mv` and `touch`, too, to avoid copying and deleting.

purcell:
I wasn't seeing this problem, but I was still seeing some misbehaviour in which old ("intermediate") flycheck results persisted even when a further change was made in quick succession, so I just recently pushed changes to address that, in 453d6d64c88df8ada4d6b9e4004c392d219e3799.

Since pushing that, I haven't noticed any issues, with the minor exception that `intero-restart` either doesn't trigger an immediate re-check or doesn't invalidate the old results.

Since I'm not really seeing an issue locally, I don't have a strong opinion on the proposed fix: the queuing makes a lot of sense, but I'm quite :-1: on sleeping for a second anywhere. I'd say if this patch is working great for you, maybe just go ahead and merge it.

chrisdone:
> I wasn't seeing this problem, but I was still seeing some misbehaviour in which old ("intermediate") flycheck results persisted even when a further change was made in quick succession, so I just recently pushed changes to address that, in 453d6d6. 

Right, I made sure to pull and recompile the latest master before embarking on this change and the GIF above is the behavior I got.

> Since I'm not really seeing an issue locally, I don't have a strong opinion on the proposed fix: the queuing makes a lot of sense, but I'm quite 👎 on sleeping for a second anywhere. I'd say if this patch is working great for you, maybe just go ahead and merge it.

Fair enough -- I don't like the sleeping either. However, the fact is that GHC's loading API is incapable of giving you sub-second subsequent turn around time. If you request "check now", then you get an immediate check. If you request "check now" and then "check again" in under a second, you get nothing for the latter, having only out of date results. So I believe the sleep is accurately modeling GHC's behavior. 

The other possible workaround would be re-implementing `:load` itself, which I'll have to look into how involved that is. But other than this subsecond issue it's quite a reliable command, so I've tried to re-use it as much as possible.

chrisdone:
Hm, I may have another much simpler solution. I'll open a separate PR for discussion.

chrisdone:
Looks super involved to rewrite :load. 

I think I'll just add a `:move` command and `:sleep` to support this solution and then this is done and dusted. It's not great to wait one second between changes but it always returns the right result, and that's what developers care about.

purcell:
Sounds good.

chrisdone:
@purcell Implemented. Do you want to give it a try on your system? Open a .hs file within the intero project itself and it'll auto-install the `0.1.22` version from the local directory. You should be able to reproduce that:

1. Making a single change returns results "immediately" (depending on size of file as usual).
2. Making changes in quick succession has a 1 second delay before getting the new results.
3. You can hit RET a few times to do the "quick succession" test.

purcell:
Cool, seems to work okay for me.

chrisdone:
Merged, intero-0.1.22 is already up on Hackage and Stackage index.

purcell:
LOVE YOUR WORK

PR #469, opened 2017-09-07
Document that `C-c C-k` quits the suggestion pane

ntc2:
We used to be able to quit the suggestion pane with `C-g`, but that
was removed in 397e5c8258c97be8e2abb1d34526219fc816ca86. The the `C-g`
was intuitive to me, but `C-c C-k` was not -- I had to look at the Git
log to figure it out -- so I thought it would be helpful to document
it. My next guess after `C-g` was `q`, which is also more intuitive
than `C-c C-k` to me.

I did not test this change, since I don't know how to run Emacs
packages from source.

### DIFF
--- a/elisp/intero.el
+++ b/elisp/intero.el
@@ -2396,7 +2396,7 @@ Each option is a plist of (:key :default :title) wherein:
                        (lambda (&rest ignore)
                          (abort-recursive-edit))
                        "C-c C-k")
-        (widget-insert (propertize " to quit without applying choices.\n\n" 'face 'font-lock-comment-face))
+        (widget-insert (propertize " to cancel.\n\n" 'face 'font-lock-comment-face))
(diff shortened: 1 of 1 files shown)


purcell:
I've changed my mind about the `C-g` in this special case, so I've just restored that binding. I'll commit a slightly-edited version of this PR - thanks!

chrisdone:
I think C-g in this case is pretty good as multiswitch is similar to isearch in its focus-stealing behavior.

purcell:
> I think C-g in this case is pretty good as multiswitch is similar to isearch in its focus-stealing behavior.

Agree. The Elisp manual says that bindings ending with `C-g` are disallowed, and `package-lint` enforced this, but there's prior art for minibuffer-style input (like `magit`'s popups) so I've softened the check and reverted this particular case.

PR #472, opened 2017-09-09
Remove old reference to intero-check-calculate-hash

### LINKED ISSUE #471: Error while checking syntax automatically: (void-function intero-check-calculate-hash) [6 times]
Getting the above error after the 1.22 release.

This is because `intero-check-calculate-hash` was removed in #468.

joehillen:
Oh the joys of life without a type-checker!

Fixes #471


purcell:
Thanks @joehillen - just encountered this myself!

PR #483, opened 2017-10-23
Add `defcustom` to disable `pop-to-repl` on `C-c C-l`.

the-kenny:
This pull request adds the option to disable `pop-to-buffer` when evaluating code either via `intero-repl-eval-region` or `intero-repl-load` (or any other function that uses `intero-with-repl-buffer`). This was done purely out of personal need as my other REPL environment (clojure / cider) doesn't pop up the repl buffer on evaluate :-) 

One could also implement a more sophisticated approach, either allowing full customization of what function is used for showing the buffer, or by allowing users to choose if they want to jump to the buffer based on some other predicate, for example the prefix-arg. 

This minimal approach is just a first step (and to probe if you're interested in this kind of improvements to `intero-mode`).

Cheers!

### DIFF
--- a/elisp/intero.el
+++ b/elisp/intero.el
@@ -131,8 +131,7 @@ To use this, use the following mode hook:
 
 (defcustom intero-pop-to-repl
   t
-  "Wether to call `pop-to-buffer' when code is sent to the
-  REPL."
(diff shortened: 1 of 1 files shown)


purcell:
Thanks. I fixed the docstring and merged this.

Issue #476, opened 2017-10-04
"Couldn't guess that module name." error in Vs-code and Haskero


deflexor:
If source file compiles with errors then type signature doesn't work, intead type
definitions of everything in this file look like:
`foldl Couldn't guess that module name. Does it exist?`

CaptJakk:
This is extremely aggravating. I am having the same issue.

purcell:
> If source file compiles with errors then type signature doesn't work

I can understand that this is frustrating, but the above could be re-phrased as "when the Haskell compiler cannot figure out the types of my definitions, `intero` won't tell me their types." In other words, if you understand that Intero is delegating this understanding of the code to GHCi, then this is the expected result by default. :-)

So there are hard limits on what `intero` can do to help. Let's consider a couple of options.

In the case where the code is well-formed but not well-typed, `intero` would theoretically be able to re-run compilation with `-fdefer-type-errors`, which would allow it to return _some_ types at the expense of a lot of extra compilation, latency and complexity.

Alternatively, `intero` could run "green"/"blue" backends, and return type information from the backend in which the file had last compiled cleanly. You can imagine how quickly this info could get out of sync with the edited file, though, and also how much complexity this approach would add.

So I don't want to answer on behalf of all the other `intero` maintainers, but IMO this is extremely unlikely to get "fixed".

chrisdone:
There's actually something possibly more hopeful and reliable possible today! ☝️ 😄 

Previously `-fdefer-type-errors` was impractical because it classed all type errors as warnings, without a way to differentiate. 

On GHC 8, they changed their error messaging system so that it became possible to differentiate these, but I was waiting a while for GHC 8 to become established before taking advantage of this fact. Now that it's more available, we, I believe, could enable `-fdefer-type-errors` on Intero start up.

![screenflow](https://user-images.githubusercontent.com/11019/33479994-1ea9a7ee-d687-11e7-8bce-672baf225801.gif)

Your editor (in this case Emacs is already setup to do it [here](https://github.com/commercialhaskell/intero/blob/648903c6d82b061bcc55dbed67f7f585027613b9/elisp/intero.el#L740..L742), which you can ask the maintainers of Haskero or Vs-Code to implement to recognize these warnings as errors, and enable `-fdefer-type-errors` when you start the project. I suggest opening an issue to them linking to here.

I'll open a separate issue/PR for the Emacs Lisp binding. It probably needs to be trialled by someone for a day. I don't mind trialling it.

[...]

PR #495, opened 2017-12-01
Enable -fdefer-type-errors on start-up

chrisdone:
See https://github.com/commercialhaskell/intero/issues/476#issuecomment-348467874 for detailed explanation.

I'm going to spend the day on this branch; if it doesn't cause any regressions in behavior (or bad performance), I'll merge to master.

If you want to test this out now, just run `killall intero` and start the processes up again.

### DIFF
--- a/elisp/intero.el
+++ b/elisp/intero.el
@@ -726,7 +726,8 @@ CHECKER and BUFFER are added to each item parsed from STRING."
     (insert string)
     (goto-char (point-min))
     (let ((messages (list))
-          (temp-file (intero-temp-file-name buffer)))
+          (temp-file (intero-temp-file-name buffer))
+          (found-error-as-warning nil))
       (while (search-forward-regexp
               (concat "[\r\n]\\([A-Z]?:?[^ \r\n:][^:\n\r]+\\):\\([0-9()-:]+\\):"
                       "[ \n\r]+\\([[:unibyte:][:nonascii:]]+?\\)\n[^ ]")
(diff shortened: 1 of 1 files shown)


chrisdone:
@purcell Are you aware of any performance numbers showing that this is not usable? I'm going to test it on a few real world big files today and report if it's a bad experience.

chrisdone:
Loading `Types.hs` from my Duet project (654 lines of types and derivings):

|Normal|With deferred type errors|
|:---|:--|
|13.658 seconds| 13.329 seconds|

No actual observed difference here.

chrisdone:
From `Server.hs` from a client project, 

|Situation|Normal|With deferred type errors|
|:--|:--|:--|
|Type error|0.354 seconds|0.813 seconds|
|No type error|0.477 seconds|0.515 seconds|

So the numbers are slower when there's more code going on. Flagging up a type error is twice as slow, and flagging that the code is OK takes an extra 100ms.

chrisdone:
It seems to vary based on caching or garbage collection inside the GHC API, because it goes as low as 

    0.643 seconds

So, really, roughly twice as slow at picking up type errors. You basically pay the price of doing a full type check every time you request a `:load`, whether it's a type error or not.

purcell:
I feel like it's probably workable. It's a little slower for me on some files which were often annoyingly slow to flycheck anyway.

elisp/intero.el
```diff
@@ -737,8 +738,10 @@ CHECKER and BUFFER are added to each item parsed from STRING."
                (msg (match-string 3)) ;; Replace gross bullet points.
                (type (cond ((string-match "^Warning:" msg)
                             (setq msg (replace-regexp-in-string "^Warning: *" "" msg))
-                            (if (string-match "^\\[-Wdeferred-type-errors\\]" msg)
-                                'error
+                            (if (or (string-match "^\\[-Wdeferred-type-errors\\]" msg)
```
purcell: You can just use `string-prefix-p` here.

chrisdone:
In the interest of always "working" (or, working more often), I think I'll merge. 

The speed hit does bother me a bit, but we can always be cleverer later to optimize the speed.

purcell:
After using this for a day or two, I reckon this has actually made the common case of flycheck error reporting significantly worse: if you refer to an unimported function, for example, then deferred errors will be reported in the code which calls it, but the root problem itself will not be detected. I feel like reverting this might be the best course of action in the first time.

chrisdone:
@purcell I think I pushed a change to fix that - can you try latest master? 

I noticed the same problem; out of scope identifiers become typed holes with deferred type errors on.

Can you still reproduce your problem?

purcell:
Yeah, I _think_ it's better now. Will plug away and report back. :-)

chrisdone:
@purcell Cool!

PR #496, opened 2017-12-05
Handle a null (buffer-file-name) when white/black listing

acowley:
I ran into trouble when exporting some Org files that had Haskell ~src~ blocks since I switched to white-listing.


chrisdone:
Thanks, I've noticed this too but didn't look into it.

PR #497, opened 2017-12-07
Do not sleep-for in with-current-buffer

CthulhuDen:
Deferred executors will run in wrong buffer context.

### DIFF
--- a/elisp/intero.el
+++ b/elisp/intero.el
@@ -1889,8 +1889,8 @@ yaml config to use, or stack's default when nil."
      result
      (lambda (result reply)
        (setf (car result) reply)))
-    (with-current-buffer (intero-buffer worker)
-      (while (not (null intero-callbacks))
+    (let ((buffer (intero-buffer worker)))
+      (while (not (null (buffer-local-value 'intero-callbacks buffer)))
         (sleep-for 0.0001)))
     (car result)))
 


chrisdone:
`sleep-for` is a blocking procedure, it's not deferred. There's no way to "cancel" it.

CthulhuDen:
I've probably messed a few terms due to my unfamiliarity with emacs. What I meant was other code in background (not sure how it was initialized, I thought it was `run-with-idle-timer`, but I experimented and it seems idle timer does not trigger during sleep) runs in wrong buffer context when intero is sleeping. I'll investigate more but I can tell you that there is observable difference between the code how it is in master and my changes.

chrisdone:
How does the following behave on your Emacs?

```lisp
     (progn
       (defvar continue)
       (setq continue t)
       (run-with-timer 5 nil
                       (lambda ()
                         (switch-to-buffer (get-buffer-create "bar"))
                         (message "Switched buffer to: %S" (current-buffer))
                         (setq continue nil)))
       (with-current-buffer (get-buffer-create "foo")
         (while continue
           (message "Sleeping ...")
           (sleep-for 1))
         (message "Final buffer is: %S" (current-buffer))))
```

```
     Sleeping ... [5 times]
     Switched buffer to: #<buffer bar>
     Final buffer is: #<buffer foo>
     "Final buffer is: #<buffer foo>"
```

CthulhuDen:
Same:

```
Sleeping ... [5 times]
Switched buffer to: #<buffer bar>
Final buffer is: #<buffer foo>
```

CthulhuDen:
Running the following:
```lisp
  (with-current-buffer (get-buffer-create "foo")
    (progn
      (defvar continue)
      (setq continue t)
      (run-with-timer 5 nil
                      (lambda ()
                        (message "Buffer is %S" (current-buffer))
                        (setq continue nil)))
      (with-current-buffer (get-buffer-create "bar")
        (while continue
          (message "Sleeping ...")
          (sleep-for 1)))
      (message "Final buffer is: %S" (current-buffer))))
```
i get this:
```
Sleeping ... [5 times]
Buffer is #<buffer bar>
Final buffer is: #<buffer foo>
```
As an example of sleeping inside with-current-buffer affecting other code

chrisdone:
That's because `sleep-for` handles the next event which is a timer event and it inherits the current buffer. Timers should set the buffer they want to be run in, otherwise there's no guarantee which buffer they'll run in, as it depends on what function handles the event loop. That's what I do here: https://github.com/commercialhaskell/intero/blob/2499ec18b1f12f430bfe950aacf21bef6ee1fe58/elisp/intero.el#L1983..L1986

So if you have some timers or code that uses them, I'd suggest fixing them!

But seeing as this doesn't change anything for intero, I'll merge.

Issue #474, opened 2017-09-14
Keybinding for :r in Intero


saurabhnanda:
When using `stack ghci` on the terminal, I notice the following behaviour:

*  `:l` to loads a module, and compiles all the dependencies of that module first. 
* If I hit `:l` again, it loads/compiles **everything** all over again.
* However, if I hit `:r` it only compiles the final module.

What is the equivalent of `:r` in Intero? I have a workaround which involves doing `:set -fobject-code` in the intero REPL, but it exhibits some weird behaviour about what modules are in-scope in the REPL.

chrisdone:
I think for a given REPL we could have a local variable of "last loaded module". If the last loaded module is the same as the one currently requested, we could run `:r` instead of `:l`. 

I don't *think* there are any problems with that.

chrisdone:
There *might* be problems if the module wasn't successfully loaded -- I suspect GHCi won't do a ":reload" without a successful ":load". We'd need to confirm that first.

khanage:
Just testing this out - if I load a module with a parse error, `:r` ghci will try to reload that same module, so no issue there.

Looking at [the elisp here](https://github.com/commercialhaskell/intero/blob/2d3c04ce523674e79f9339189a79e6841e1fa587/elisp/intero.el#L1115) looks like it would be simplest to do on a per-frontend basis (as the string ":load File.hs" gets passed directly to the backend?

kryoxide:
I think the merged pull request #498 resolves this issue?

PR #498, opened 2017-12-08
#474 Only ":reload" file if it is equal to "intero-repl-last-loaded"

kryoxide:
This checks if the path of the current file is equal to the content of
"intero-repl-load". If the paths are equal, the file is only reloaded, which
skips recompilation of unchanged dependency modules.

This is faster and a bit more convenient in larger projects.

This would resolve enhancement request #474

### DIFF
--- a/elisp/intero.el
+++ b/elisp/intero.el
@@ -1133,10 +1133,16 @@ If PROMPT-OPTIONS is non-nil, prompt with an options list."
   (save-buffer)
   (let ((file (intero-localize-path (intero-buffer-file-name))))
     (intero-with-repl-buffer prompt-options
-      (comint-simple-send
-       (get-buffer-process (current-buffer))
-       (concat ":load " file))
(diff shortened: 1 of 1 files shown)


chrisdone:
Cool. I can't remember why I haven't added this before. I guess we'll find out.

kryoxide:
On issue #474 you were concerned that the ":reload" wouldn't work properly if there was a compile-error in the initial ":load".

I had this change running on my machine for some time now and got no problems so hopefully it works for everyone.

chrisdone:
This also needs to be preserved through GHCi reloads.

chrisdone:
I.e. if you restart ghci then `:r` doesn't work, and the elisp keeps sending `:r` and so the user is stuck.

chrisdone:
I have an idea for an alternative way of doing this.

chrisdone:
OK, pushed a fix for this. Now it resets the last-loaded value when you start the REPL.

PR #500, opened 2017-12-15
Revert "#474 Only ":reload" file if it is equal to "intero-repl-last-loaded""

chrisdone:
Reverts commercialhaskell/intero#498

If the module isn't successfully loaded, ghci just says:
```
 Ok, modules loaded: none.
 Ok, modules loaded: none.
 Ok, modules loaded: none.
 Ok, modules loaded: none.
```

### DIFF
--- a/elisp/intero.el
+++ b/elisp/intero.el
@@ -1133,16 +1133,10 @@ If PROMPT-OPTIONS is non-nil, prompt with an options list."
   (save-buffer)
   (let ((file (intero-localize-path (intero-buffer-file-name))))
     (intero-with-repl-buffer prompt-options
-      (if (or (not intero-repl-last-loaded)
-	      (not (equal file intero-repl-last-loaded)))
-	  (progn
-	    (comint-simple-send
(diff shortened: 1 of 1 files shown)


kryoxide:
Strange, when loading a file that doesn't compile I get output like:

```
λ [1 of 3] Compiling GhciTypes        ( /Users/kryoxide/programming/Haskell/intero/src/GhciTypes.hs, interpreted )
[2 of 3] Compiling GhciInfo         ( /Users/kryoxide/programming/Haskell/intero/src/GhciInfo.hs, interpreted )
[3 of 3] Compiling GhciFind         ( /Users/kryoxide/programming/Haskell/intero/src/GhciFind.hs, interpreted )
/Users/kryoxide/programming/Haskell/intero/src/GhciFind.hs:453:1:
    parse error (possibly incorrect indentation or mismatched brackets)
Failed, modules loaded: GhciInfo, GhciTypes.
```

And everything works just fine in my further workflow. If you have any idea what could cause the problem on your machine I will try to fix it.

Issue #324, opened 2016-11-28
Install intero once per ghc version


mgsloan:
I've opened a PR for stack adding a proposed `--copy-compiler-tool` flag.  This flag causes stack to copy executables to a dir for compiler-specific tools, e.g. `/home/mgsloan/.stack/compiler-tools/x86_64-linux/ghc-8.0.1/bin`.  The intention of this is to allow installation of tools that need to match the ghc version.

In order to use this with intero, it seems like the following will need to happen:

1) Add checking of stack version - https://github.com/commercialhaskell/intero/issues/127
2) Pass in this new flag when automatically building intero

Ideally, we'd also check that we are using an intero that matches our GHC - https://github.com/commercialhaskell/intero/issues/161 .  One option here might be to demand that intero be located at `stack path --compiler-tools-bin`

The way intero is installed could be sensitive to particular choices in the stack.yaml (dependencies, etc).  I think it also makes sense to have configurations like `stack-7.8.yaml`, `stack-7.10.yaml`, etc for each compiler version.  Then, these configurations can be used when installing intero.

Thoughts?  Perhaps this isn't worth the effort, maybe the new `--copy-compiler-tool` flag shouldn't be merged?

chrisdone:
Interesting stuff I'll look into it when I get home.

ntc2:
Just worked around a failing project-local intero install by installing globally with `--copy-compiler-tool` as suggested here:

https://github.com/commercialhaskell/intero/issues/504

If intero installed itself globally in this way, instead of in my project, that would resolve the problem for me automatically.

chrisdone:
I need to know the version of stack that introduced that flag, and then
intero can do a version detection to use that flag or not. Otherwise it'll
obviously fail if the stack version is older.

ntc2:
Looks like `--copy-compiler-tool` was added in Stack 1.6.1:
https://github.com/commercialhaskell/stack/blob/master/ChangeLog.md#v161 .

mgsloan:
Note that if `stack build --copy-compiler-tool` is used within a project, then that particular project's settings and dependencies will affect the resulting `intero` binary.  This is why in the original issue I'm describing having different stack.yaml files for building intero, based on ghc major version.

chrisdone:
@mgsloan OK, so it would have to detect the GHC version with

```
-bash-3.2$ stack ghc -- --version
The Glorious Glasgow Haskell Compilation System, version 8.0.2
```

[...]

PR #505, opened 2018-01-03
Use --copy-compiler-tool for building intero

varosi:
This will copy _Intero_ in a GHC version specific place. When you do **stack exec intero** it will run matching version _Intero_.
It'll also save build time as _Intero_ will be installed once per GHC version and not for each project.

### DIFF
--- a/TOOLING.md
+++ b/TOOLING.md
@@ -10,20 +10,20 @@ It's basically GHCi plus extra features. Those are:
 
 # Installing
 
-Use `stack build` (not `install`) for each of your package sets. Each
+Use latest [Stack build tool](https://haskellstack.org). Use `stack build` (not `install`) for each of your package sets. Each
 LTS or nightly should have a separate `stack build`. **If you use**
 `stack install` **you will run into incompatibility issues
 between package sets.**
 
 Standard:
 
-    $ stack build intero
+    $ stack build --copy-compiler-tool intero
 
 From source:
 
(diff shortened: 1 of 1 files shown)


mgsloan:
Hmm, it may well make sense to recommend to people to do something like `stack install --resolver --copy-compiler-tool intero` (not sure, but I *think* "install" instead of "build" might be needed for it to have any effect).

However, right now the behavior of that command depends on your local or global project.  I have opened https://github.com/commercialhaskell/stack/issues/3743 to resolve this.  Perhaps for now the docs could say to do something like a few different variations of `stack install intero --resolver lts-10.1 --copy-compiler-tool` for different GHCs.  Note also that the invocation should manually specify the latest version of intero.  I haven't thought through it fully for this usecase, but it would also be good to be able to install the latest version of a package with the build invocation - https://github.com/commercialhaskell/stack/issues/3744

Also, note that usually intero gets installed by the emacs config.  I have some notes here https://github.com/commercialhaskell/intero/issues/324 about ideas for doing this, namely:

> The way intero is installed could be sensitive to particular choices in the stack.yaml (dependencies, etc). I think it also makes sense to have configurations like stack-7.8.yaml, stack-7.10.yaml, etc for each compiler version. Then, these configurations can be used when installing intero.

I kinda like the idea of having this all be in a copy-pastable reproducible CLI invocation that's documented and in the elisp, though.

chrisdone:
Intero is now installed with --copy-compiler-tool automatically.

PR #515, opened 2018-01-31
Primarily customisability enhancements

sergv:
I would like propose a bunch of improvements, which can be roughly summarised as follows:
- Performance enhancements - emacs would do less work and would (hopefully) work faster. The only danger here is that when I changed `string-match` to `string-match-p` and `looking-at` to `looking-at-p` I may have missed a place where match data was subsequently accessed. But hopefully that didn't happen...
- Customisability enhancements - move more stuff (i.e. keymap for `intero-multiswitch`) to variables or add new variables (i.e. `intero-extra-ghc-options` and `indero-extra-ghci-options`) so that a user may influence intero's behaviour without having to change it's source code.
- New feature - completion in repl for `:set` command. Could probably be improved to complete other commands as well, but I'd rather add this small bit now and come up with a followup.
- Windows bug fix - `intero-paths-for-same-file` does not do the right thing on Windows platform whet fed arguments like `C:/foo/bar/Baz.hs` and `C:\foo\bar\Baz.hs`.

### DIFF
--- a/elisp/intero.el
+++ b/elisp/intero.el
@@ -360,7 +360,7 @@ You can use this to kill them or look inside."
   (interactive)
   (let ((buffers (cl-remove-if-not
                   (lambda (buffer)
-                    (string-match " intero:" (buffer-name buffer)))
+                    (string-match-p " intero:" (buffer-name buffer)))
                   (buffer-list))))
     (if buffers
         (display-buffer
@@ -534,7 +534,7 @@ If the problem persists, please report this as a bug!")))
                           (forward-char 1)
                           (looking-back "$(" 1))
                         (+ 2 (current-column))
-                      (if (looking-at "$(")
+                      (if (looking-at-p "$(")
                           (+ 3 (current-column))
                         (1+ (current-column)))))
             (expansion-msg
@@ -648,7 +648,7 @@ running context across :load/:reloads in Intero."
          ":load DevelMain"
          (current-buffer)
          (lambda (buffer reply)
-           (if (string-match "^O[Kk], modules loaded" reply)
+           (if (string-match-p "^O[Kk], modules loaded" reply)
                (intero-async-call
                 'backend
                 "DevelMain.update"
@@ -763,14 +763,14 @@ CHECKER and BUFFER are added to each item parsed from STRING."
                      (match-string 3))) ;; Replace gross bullet points.
                (type (cond ((string-match "^Warning:" msg)
                             (setq msg (replace-regexp-in-string "^Warning: *" "" msg))
-                            (if (or (string-match "^\\[-Wdeferred-type-errors\\]" msg)
-                                    (string-match "^\\[-Wdeferred-out-of-scope-variables\\]" msg)
-                                    (string-match "^\\[-Wtyped-holes\\]" msg))
+                            (if (or (string-match-p "^\\[-Wdeferred-type-errors\\]" msg)
+                                    (string-match-p "^\\[-Wdeferred-out-of-scope-variables\\]" msg)
+                                    (string-match-p "^\\[-Wtyped-holes\\]" msg))
                                 (progn (setq found-error-as-warning t)
                                        'error)
                               'warning))
-                           ((string-match "^Splicing " msg) 'splice)
-                           (t                               'error)))
+                           ((string-match-p "^Splicing " msg) 'splice)
+                           (t                                 'error)))
                (location (intero-parse-error
                           (concat local-file ":" location-raw ": x")))
                (line (plist-get location :line))
@@ -910,7 +910,7 @@ Other arguments are IGNORED."
                 (funcall cont
                          (cl-remove-if-not
                           (lambda (candidate)
-                            (string-match (concat "^" prefix) candidate))
+                            (string-match-p (concat "^" prefix) candidate))
                           intero-pragmas))
                 t)))
         (intero-get-repl-completions source-buffer prefix cont))))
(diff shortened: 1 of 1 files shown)


elisp/intero.el
```diff
@@ -920,6 +939,23 @@ considered."
           (backward-char)
           (not (looking-at-p (rx (| space line-end)))))))))
 
+(defun intero-completions-grab-ghci-command ()
+  "Grapb prefix for a ghci command like :set.
```
acowley: s/Grapb/Grab/

acowley:
Does the [switch back to using the GHCi repl](https://github.com/commercialhaskell/intero/commit/f80d503fd682defc750406ef3021091b732edffe) impact the approach to expanding autocompletion of repl commands?

acowley:
The `-p` changes look good to me, but I haven't tried it out yet.

sergv:
Regarding 'switch back to using the GHCi repl'' it seems that at some point there might be a difference between the `intero` and the `ghci` executables. However, I expect user-visible commands like `:set` to be largely compatible.

elisp/intero.el
```diff
@@ -518,7 +534,7 @@ If the problem persists, please report this as a bug!")))
                           (forward-char 1)
                           (looking-back "$(" 1))
                         (+ 2 (current-column))
-                      (if (looking-at "$(")
+                      (if (looking-at-p "$(")
```
purcell: This is certainly a reasonable change semantically, but I'd be surprised if it affects performance. After all, `looking-at-p` is defined as
```
(let ((inhibit-changing-match-data t))
    (looking-at regexp))
```

elisp/intero.el
```diff
@@ -796,19 +812,20 @@ CHECKER and BUFFER are added to each item parsed from STRING."
...
+               do (when (string-match (car regex) string)
+                    (setq span
+                          (list :file (match-string 1 string)
+                                :line (string-to-number (match-string 2 string))
+                                :col (string-to-number (match-string 4 string))
+                                :line2 (when (match-string 3 string)
+                                         (string-to-number (match-string 3 string)))
+                                :col2 (when (match-string 5 string)
+                                        (string-to-number (match-string 5 string)))))))
+      span)))
```
purcell: +1 for this change. (The whole thing could likely be written more efficiently, by breaking the loop when a regexp matches, rather than continuing to (potentially) match other regexps.)

elisp/intero.el
```diff
@@ -893,7 +910,7 @@ Other arguments are IGNORED."
                 (funcall cont
                          (cl-remove-if-not
                           (lambda (candidate)
-                            (string-match (concat "^" prefix) candidate))
+                            (string-match-p (concat "^" prefix) candidate))
```
purcell: If `prefix` isn't a regexp, which it shouldn't be, this could/should simply be `string-prefix-p`.

elisp/intero.el
```diff
@@ -904,6 +921,8 @@ If specified, MINLEN is the shortest completion which will be
 considered."
   (when (intero-completions-can-grab-prefix)
     (let ((prefix (cond
+                   ((and (eq 'intero-repl-mode major-mode)
```
purcell: Prefer `(derived-mode-p 'intero-repl-mode)` to testing for exact equality of `major-mode`.

purcell: Generally you'd put the test for applicability (i.e. the major mode test here) in the function which grabs the prefix: it would return nil if inapplicable.

elisp/intero.el
```diff
@@ -2339,6 +2377,13 @@ a list is returned instead of failing with a nil result."
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 ;; Multiselection
 
+(defvar intero-multiswitch-keymap
+  (let ((map (copy-keymap widget-keymap)))
+    (define-key map (kbd "C-c C-c") 'exit-recursive-edit)
+    (define-key map (kbd "C-c C-k") 'abort-recursive-edit)
+    (define-key map (kbd "C-g")     'abort-recursive-edit)
+    map))
+
```
purcell: Nice.

elisp/intero.el
sergv: Well, the C implementation of `looking-at` does pay attention to this variable and does more work if it's `nil`. For instance, please check https://github.com/emacs-mirror/emacs/blob/master/src/search.c#L334 (and possibly implementation of `re_match_2` as well).

elisp/intero.el
sergv: I think there's even more performant implementation: just test all three regexes at once by single `match-string`. Regexes modify the same groups for file name, line number and column - single match should populate them properly.

sergv:
I have fixed test for `intero-repl-mode` and `string-prefix-p` over `string-match-p` for pragmas. I have also added another optimization where multiple regexps are now tested in one go through single call to `string-match`.

elisp/intero.el
purcell: Thanks, great to know.

elisp/intero.el
```diff
@@ -747,14 +763,19 @@ CHECKER and BUFFER are added to each item parsed from STRING."
...
                             (setq msg (replace-regexp-in-string "^Warning: *" "" msg))
-                            (if (or (string-match "^\\[-Wdeferred-type-errors\\]" msg)
-                                    (string-match "^\\[-Wdeferred-out-of-scope-variables\\]" msg)
-                                    (string-match "^\\[-Wtyped-holes\\]" msg))
+                            (if (string-match-p
+                                 (rx bol
+                                     "["
+                                     (or "-Wdeferred-type-errors"
+                                         "-Wdeferred-out-of-scope-variables"
+                                         "-Wtyped-holes")
```
purcell: Really nice.

PR #517, opened 2018-02-09
Support asynchronous read-only commands

chrisdone:
### This PR adds

* To the Intero executable: 
   1. The ability for read-only commands (`:type-at`, `:loc-at`, `:complete-at`, `:uses-at`) to be **called in a separate thread** to the main thread, which may be compiling a module. List [here](https://github.com/commercialhaskell/intero/blob/76ba40109ada211d76bf435ea6cb8c16227d68d5/src/InteractiveUI.hs#L301..L308). 
   2. A simple TCP service that listens on a _random available port_ on loopback (127.0.0.1-only), which is reported when it starts up as `Intero-Service-Port`.
* For the Emacs mode, 
    1. We pick up the `Intero-Service-Port`  when starting the Intero process. 
    2. We add two new functions `intero-async-network-call` and `intero-blocking-network-call` which both connect to the service. _One new connection per command request_, and after the response is sent the connection is closed. If the connection cannot be established, we fall back to talking to the intero pipe as usual.

### Why?

Because when flycheck is running which calls `:load` on the current module, if your module is big, you'll have to wait until the check is done before any other commands (listed above) can run. The intero process often already has this information, but the command queue can only run one at a time.

This makes the user experience much faster. You can edit code and get goto-definition, type info, completions, much faster than before.

### Demonstration

**Before**, notice how you have to wait for the type-check to complete (see the `FlyC*` at the bottom right) before the completion is able to trigger (we could change the hook order, but that actually doesn't improve anything):

screenflow

**After**, notice that the reply is immediate even though we're currently compiling the module. See the `FlyC` at the bottom right change.

screenflow-fast

## Testing

I'm going to work for the rest of the day with this branch and generally be aware of any misbehavior.
Meanwhile the CI will check that the changes to the intero executable build properly.

### DIFF
--- a/elisp/intero.el
+++ b/elisp/intero.el
@@ -260,6 +260,21 @@ and blacklist match, then the whitelist entry wins, and
   "List of callbacks waiting for output.
 LIST is a FIFO.")
 
+(defvar-local intero-async-network-cmd nil
+  "Command to send to the async network process when we connect.")
+
+(defvar-local intero-async-network-connected nil
+  "Did we successfully connect to the intero service?")
+
+(defvar-local intero-async-network-state nil
+  "State to pass to the callback once we get a response.")
+
+(defvar-local intero-async-network-worker nil
+  "The worker we're associated with.")
+
+(defvar-local intero-async-network-callback nil
+  "Callback to call when the connection is closed.")
+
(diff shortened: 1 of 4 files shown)


purcell:
Neato!

PR #519, opened 2018-02-10
Check if UseLoopBack socket option is supported

FilWisher:
Not all systems support the UseLoopBack socket option.
Network.Socket.setSocketOption throws an exception when a system doesn't
support the socket option. This commit checks if the system supports
`UseLoopBack` before attempting to set it.

Fixes https://github.com/commercialhaskell/intero/issues/518

### DIFF
--- a/src/InteractiveUI.hs
+++ b/src/InteractiveUI.hs
@@ -678,7 +678,10 @@ listenOnLoopback = do
     close
     (\sock -> do
        setSocketOption sock ReuseAddr 1
-       setSocketOption sock UseLoopBack 1
+       if isSupportedSocketOption UseLoopBack then
+         setSocketOption sock UseLoopBack 1
+       else
+         return ()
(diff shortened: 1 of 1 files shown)


chrisdone:
Cool! Thanks.

chrisdone:
@FilWisher BTW, which systems don't support loopback? Linux, OS X or Windows? Do I need to push a fresh release, is this urgent?

FilWisher:
AFAIK, it isn't supported on Linux. It may be worth a fresh release

chrisdone:
I'll try it on Linux, hold my beer.

chrisdone:
I'll just drop it, as I don't think it was necessary anyway with binding to 127.0.0.1.

GodPlaysChess:
Had such issue on windows 7.  Upgrading to intero-0.1.26 helped. Thanks!

chrisdone:
Sweet, thanks!

PR #522, opened 2018-02-12
Cache calls to 'stack path'

sergv:
This is useful for Windows folks since 'stack path' can take noticeable amount of time even for a single buffer thanks to antivirus ~malware~ software. But the decision to cache may be controversial as it may store old state of the filesystem. Even if there would be a handle to reset cache, users will have to be educated about it and will have to remember to use it if something does not work as intended.

I certainy feel that caching would be beneficial for Windows users (of which I am a member of), but am aware of the drawbacks. Do you think this change could lead to unintended issues too big to justify the benefits?

### DIFF
--- a/elisp/intero.el
+++ b/elisp/intero.el
@@ -2478,23 +2478,25 @@ Uses the directory of the current buffer for context."
             " "
             root)))
 
-(defun intero-project-root ()
-  "Get the current stack config directory.
-This is the directory where the file specified in
-`intero-stack-yaml' is located, or if nil then the directory
-where stack.yaml is placed for this project, or the global one if
-no such project-specific config exists."
-  (if intero-project-root
-      intero-project-root
-    (let ((stack-yaml intero-stack-yaml))
-      (setq intero-project-root
-            (intero-with-temp-buffer
+(defvar intero--project-root-cache (make-hash-table :test 'equal)
+  "Cache for `intero--get-project-root'. Maps pairs of `default-directory' and
+optional stack-yaml file to resolved project roots.")
+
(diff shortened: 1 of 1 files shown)


mgsloan:
How about invalidating the cache of `stack path` whenever the stack configuration changes?

sergv:
That may be the solution. But how to detect when stack configuration changed? Should Emacs use a hash of `stack.yaml`?

mgsloan:
Modification timestamps should be sufficient.

purcell:
@sergv Is it really necessary to check the file size?

sergv:
@purcell Well, I'd rather check it just to be sure. Do you think it's redundant in all conceivable circumstances?

purcell:
> Well, I'd rather check it just to be sure. Do you think it's redundant in all conceivable circumstances?

I think anything short of hashing the file contents is going to have edge cases, so mod time is simple and sufficient. (P.S. There's the `buffer-hash` function, but it's not available in all our supported Emacs versions, and not worth reimplementing locally.)

sergv:
Any update on this, please?

sergv:
I just looked more closely and my submissions are total bollocks! Just the fact that `locate-dominating-file` returns a directory, not the `stack.yaml` file, is enough to make all this caching facility broken beyound repair. But then, if that gets fixed, it looks like calling `stack path` will be rendundant since it's not going to do anything more sophisticated than look up a directory with a `stack.yaml` file, which will be already performed by the elisp. I'll close this for now, lest it gets merged and thus adds my embarassing commits to the project's history.

PR #523, opened 2018-02-12
Delete network buffers after use

sergv:
I noticed that a new network buffer is created for each 'read-only' command like get type of thing at point. After a while a lot of such buffers is created but they are never cleaned up.

### DIFF
--- a/elisp/intero.el
+++ b/elisp/intero.el
@@ -2015,15 +2015,22 @@ as (CALLBACK STATE REPLY)."
       (error "Intero process is not running: run M-x intero-restart to start it"))))
 
 (defun intero-network-call-sentinel (process event)
-  (with-current-buffer (process-buffer process)
+  (let* ((buf (process-buffer process))
+         (cleanup
+          (lambda ()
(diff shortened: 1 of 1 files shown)


elisp/intero.el
```diff
@@ -2015,15 +2015,22 @@ as (CALLBACK STATE REPLY)."
       (error "Intero process is not running: run M-x intero-restart to start it"))))
 
 (defun intero-network-call-sentinel (process event)
-  (with-current-buffer (process-buffer process)
+  (let* ((buf (process-buffer process))
+         (cleanup
+          (lambda ()
+            (delete-process process)
```
purcell: I think that if you use `cl-flet` here instead, you wouldn't need to `funcall` the `cleanup` function. Looks reasonable otherwise.

sergv:
I'd rather prefer plain `funcall` due to its simplicity, but `cl-letf` seems reasonable too. Fixed.

purcell:
Great, thanks!

Issue #524, opened 2018-02-15
RFC: hole filling


chrisdone:
I've implemented a working hole filler on a branch in the `intero` executable. The motivation is to fill in functions that are obvious from the types, or that you know exists but you can't remember the name exactly, or you don't feel like typing it. 

This means that you can write code like this:

```haskell
import qualified Data.Map.Strict as M
import           Data.Generics as K
main :: IO ()
main = _ (M.insert 'a' 'b' _)
```
And run `:fill Demo 4` (fill in the module Demo the declaration on line 3) and get:

```haskell
Completion: print, mempty
Completion: print, K.empty
Completion: print, M.empty
Completion: print, undefined
Completion: mempty, mempty
Completion: mempty, K.empty
-- etc.
```

All these completions are well-typed and work together. Something simpler:

``` haskell
main :: IO ()
main = putStrLn (_ 123)
```
```
Completion: show
Completion: mempty
Completion: undefined
```
An example from within the source that provides this feature:

```haskell
    unambiguous grelt = null conflicts
      where
        conflicts =
          filter
            greltUnqualified
            (filter
               (/= grelt)
               (lookupGlobalRdrEnv gre (_ (gre_name grelt))))
```
```haskell
Completion: getOccName
Completion: occName
Completion: nameOccName
Completion: Data.Generics.empty
Completion: undefined
Completion: errorBadArgument
Completion: errorMissingArgument
Completion: errorShortFormat
```

I'm wondering how to implement a UI for this in Emacs.

*  One option is to just re-use the suggestions system of Intero, listing all the combinations: x,y,z, x,y,c, x,k,m etc. 
* Another would be to show an overlay on the first hole e.g. `[print]` and then you can somehow UI-toggle other options. Upon committing to one choice from that hole, it moves onto the next one. That permits you to interactively narrow down things if you like. 
* Just show one hole completion at a time in Intero's suggestions system.
* With a keybinding, manually trigger company-mode to present a list on the first hole. And repeat. I kind of like this idea. We could also make simply `_` trigger company-mode in this special way. If I can just figure out how to replace the `_` hole entirely after choosing a choice, this might be ideal!

Completion for one hole with a big set of imports (e.g. lots of GHC imports) with 5K~ symbols takes around 200ms. For more holes, it can explode to 10 seconds. So it may be best to go one hole at a time.

Any input from Intero users welcome.

david-christiansen:
[...]

PR #526, opened 2018-02-18
Add :fill command

chrisdone:
As discussed in #524. In order to aid experimenting, I'm merging this into master once the CI checks come through.

This feature will only trigger if you type `_` and have company-mode enabled (which is intero default). Alternatively, if you run `M-x company-complete` (or other keybinding). I've rebound my `M-/` to dabbrev-expand unless on a `_` and then it runs `company-complete`.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,3 +1,6 @@
+0.1.27:
+	* Add experimental :fill command.
+
 0.1.23:
 	* :cd handles quoted filepaths properly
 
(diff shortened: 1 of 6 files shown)


chrisdone:
Seems that this doesn't build on GHC 8.0.2 despite the CI confirmation... have to fix that.

chrisdone:
Fixed the CI scripts, now it should fail properly. Let's see.

PR #528, opened 2018-02-20
Revert "Delete network buffers after use"

purcell:
Reverts commercialhaskell/intero#523

With the above change, I see a lot of:

```
Debugger entered--Lisp error: (error "Selecting deleted buffer")
  intero-network-call-sentinel(#<process backend> "deleted\n")
```

so it seems like this fix isn't the correct one.

/cc @sergv

### DIFF
--- a/elisp/intero.el
+++ b/elisp/intero.el
@@ -2065,39 +2065,33 @@ as (CALLBACK STATE REPLY)."
       (error "Intero process is not running: run M-x intero-restart to start it"))))
 
 (defun intero-network-call-sentinel (process event)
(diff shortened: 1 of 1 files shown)


sergv:
Is there any way to reproduce the problem?

sergv:
It's very strange - I cannot get `intero-network-call-sentinel` to receive "deleted\n" event. All I'm getting locally is "open\n" (on initialisation) or "connection broken by remote peer\n" (on finish). Perhaps "deleted\n" event may be safely ignored?

Issue #529, opened 2018-02-24
Intero REPL autocomplete problem


joswik:
Selecting a Language extension in the REPL from dropdown menu removes all text before it:
e.g: 
λ :set -XOv (RET)
  you get this:
λ -XOverloadedStrings

It doesn't happen in a regular haskell source code buffer 
i.e. {-# LANGUAGE OverloadedStrings #-} autocompletes fine.

Tested with the latest intero (20180221.149),
Emacs 25.2, GHC 8.0.2 (Win7 32-bit), and
Emacs 25.3, GHC 8.2.2 (Mac 10.12.6) - with only intero and stack installed.

chrisdone:
Doesn't reproduce for me, must've been fixed by #530.

joswik:
Yes, it's  been working correctly for a long time.

Thanks
Josko
________________________________

PR #530, opened 2018-02-25
ghci-like repl autocomplete

KarlFish:
I created the function  `intero-repl-completion-at-point`, an alternative company function. This should provide a native ghci auto-complete experience in the intero repl when paired with the changes from https://github.com/commercialhaskell/intero/commit/f80d503fd682defc750406ef3021091b732edffe. 

Here's a small demo of a user experience before and after. `After` includes the `notInExport` in completions and fixes issue https://github.com/commercialhaskell/intero/issues/529 

Before

After

### DIFF
--- a/elisp/intero.el
+++ b/elisp/intero.el
@@ -871,6 +871,30 @@ CHECKER and BUFFER are added to each item parsed from STRING."
           (when completions
             (list beg end completions)))))))
 
+(defun intero-repl-completion-at-point ()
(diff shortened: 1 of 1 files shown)


chrisdone:
Nice!

ntc2:
Could someone explain how enable this ` intero-repl-completion-at-point` completion in the REPL? I'd be happy to update the README once I understand how to use this.

ntc2:
Please disregard last comment. This is already working for me out of the box :)

PR #535, opened 2018-03-07
Add missing field suggestions

chrisdone:
screenflow


mgsloan:
Pretty cool!

I realize it is probably easier to write this stuff in elisp, but it would be nice if such logic was instead written in Haskell so that this can be shared with editors other than emacs.

PR #541, opened 2018-03-31
Allow spaces in file paths

milapsheth:
Intero currently runs commands that don't work well with paths containing spaces,
eg. `:load C:\Program Files\src\Hello.hs`.
This fix adds quotes around paths.

### DIFF
--- a/elisp/intero.el
+++ b/elisp/intero.el
@@ -550,7 +550,7 @@ If the problem persists, please report this as a bug!")))
     (intero-with-dump-splices
      (let* ((output (intero-blocking-call
                      'backend
-                     (concat ":load " (intero-localize-path (intero-temp-file-name)))))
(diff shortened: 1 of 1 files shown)


purcell:
Thanks @milapsheth - I agree this needs fixing, and I've tackled it a slightly different way which I believe will be safer. Please give the code in #542 a spin and let me know if it works for you on Windows.

In particular, you'll note that the code in your change which included `(format "-odir=%S" dir)` is not present in my changes: I believe that particular `%S` adds one level of quoting too many, and will lead to problems. Perhaps you can confirm that it is not necessary?

milapsheth:
Sweet, thanks for the prompt fix! I was afraid to push a non-trivial change since I'm not familiar with the code so I agree that #542 is a better fix. I also noticed something minor and did a review.
So I'll close this pull request in favour of #542.

PR #542, opened 2018-03-31
Properly and consistently quote file paths for GHCi :commands

purcell:
- Quote all paths with ""
- Escape only \ and "
- Don't rely on %S elisp string escaping
- Remove special-case handling for :move

This allows all commands to safely handle Windows paths and both directory paths and filenames which contain spaces on any platform.

This should replace #541 if it is found to work for all concerned - I've tested this on a UNIX machine to my satisfaction.

### DIFF
--- a/elisp/intero.el
+++ b/elisp/intero.el
@@ -546,7 +546,7 @@ If the problem persists, please report this as a bug!")))
     (intero-with-dump-splices
      (let* ((output (intero-blocking-call
                      'backend
-                     (concat ":load " (intero-localize-path (intero-temp-file-name)))))
+                     (concat ":load " (intero-path-for-ghci (intero-temp-file-name)))))
             (msgs (intero-parse-errors-warnings-splices nil (current-buffer) output))
             (line (line-number-at-pos))
             (column (if (save-excursion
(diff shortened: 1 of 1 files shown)


elisp/intero.el
```diff
@@ -1796,7 +1795,7 @@ type as arguments."
 (defun intero-format-get-type-at (beg end)
   "Compose a request for getting types in region from BEG to END."
   (format ":type-at %S %d %d %d %d %S"
-          (intero-localize-path (intero-temp-file-name))
+          (intero-path-for-ghci (intero-temp-file-name))
```
milapsheth: I believe that the format string '%S' here should also be changed to '%s'.

milapsheth:
I can confirm that this change fixes the issue on my local Windows 10 machine.

elisp/intero.el
purcell: Indeed, thanks - this is now fixed.

purcell:
> I can confirm that this change fixes the issue on my local Windows 10 machine. The extra quotation for -odir=%s was not needed.

Thanks for the confirmation, and for testing all of this locally. Since it works for both of us, I think I'll go ahead and merge this shortly, despite having pinged Chris for a review.

milapsheth:
@purcell So I looked a bit more into the `-odir` issue. It turns out that it extra quoting fails when creating options for `intero` with exception `option --ghci-options: unterminated string: endOfInput` but works for `ghci`. In the latter case of `ghci`, not quoting it gives an error only in a local project but I don't fully understand the consequence of the error (It only complains that `target 'path_with_spaces' is not a module or source file.

purcell:
> It might make sense to fix it by using intero-path-for-ghci when options for ghci are created and intero-localize-path when options for intero are created although I'm not sure if there's a deeper issue here.

I don't really understand what you mean by "ghci" vs "intero" here, sorry.

milapsheth:
> I don't really understand what you mean by "ghci" vs "intero" here, sorry.

I should have clarified. I meant argument passed to `intero-make-options-list`. There are two places where this function is called; once with argument `ghci` and the other with `intero`.

purcell:
Ah, thanks. I hadn't made that connection. I'll take a look.

milapsheth:
To clarify a bit more:
When I start a repl where the path to `-odir` contains spaces, it complains that
`target 'path_with_spaces' is not a module name or a source file`
but still runs the repl and loads the file correctly. So, I'm not sure how the error manifests itself but it does complain about it (even when run in a separate shell). 
If I run the command in a shell and add extra quotes to the argument of `-odir=`, then it doesn't complain. But if I run it in emacs with extra quotes added, then it gives me the error
`option --ghci-options: unterminated string: endOfInput`.

purcell:
The quotes in a comment when you run it in the shell won't behave exactly the same as when they are added to elements of an argument list in Emacs, since that list is passed as an array directly to the `execv` system call, and not executed by your shell. So the Emacs command in this case is not evaluated by your shell, and quotes that would be necessary for your shell to distinguish separate arguments will not be necessary when those arguments are passed directly to the OS.

Given the above, where you're seeing a spurious error but are still able to continue, and given that this PR hasn't changed that specific chunk of code, I'll merge this now anyway, but I'd be keen to hear any thoughts @chrisdone might have on what you're seeing.

milapsheth:
I see, that makes a lot of sense. Thanks a lot for fixing the main issue. It seems that the other error will be triggered when `.hi` or `.o` files are generated and won't be put in the corresponding `.hidir` and `.odir` directory. But these files aren't generated when running the command for the repl.

Issue #544, opened 2018-04-05
Many ` intero-network:backend-` buffers getting created


ChickenProp:
As I use intero, it seems that the length of my buffer list keeps increasing. This is a problem for me, as it slows down `ido-switch-buffer`.

It looks as though the problem is caused by a lot of buffers named like ` intero-network:backend-*` (with a leading space). For example, if I do `(insert (substring (format "%S" (mapcar 'buffer-name (buffer-list (selected-frame)))) 0 1000))`, I get

```
("*scratch*" " *Minibuf-1*" "*Backtrace*" "*Messages*" "PermissionSpec.hs" "TestQueryUtil.hs" "Api.hs" "Review/Queries.hs" "src/Queries.hs" "Standardize.hs" "excelsior.cabal" "QuerySpec.hs" "Util.hs" "CDMVarIds.hs" "User/Types.hs" "Reports/Queries.hs" " *Minibuf-0*" "Properties/Queries.hs" "Review/Types.hs" ".dir-locals.el" "*GNU Emacs*" " *code-conversion-work*" " *Echo Area 0*" " *Echo Area 1*" " intero-network:backend" " intero-network:backend-150056" " intero-network:backend-806060" " intero-network:backend-958468" " intero-network:backend-695967" " intero-network:backend-359141" " intero:backend:excelsior /home/phil/code/excelsior" " intero-network:backend-241704" " intero-network:backend-804390" " intero-network:backend-853775" "*Help*" " intero-network:backend-655359" " intero-network:backend-644620" " intero-network:backend-119029" " intero-network:backend-537428" " intero-network:backend-262984" " intero-network:backend-706094" " intero-network:backend-799389" " intero-network
```

Yesterday I restarted emacs to clear these out (`intero-restart` didn't seem to do the trick), but today my buffer list was already over 3000 items. Just now I did

```elisp
(defun intero-buffers (frame)
  (delq nil
	(mapcar
	 (lambda (x)
	   (let ((name (buffer-name x)))
             (if (string-prefix-p " intero-network:backend-" name)
		 name)))
	 (buffer-list frame))))

(defun kill-intero-buffers ()
  (mapcar 'kill-buffer (intero-buffers (selected-frame))))
(kill-intero-buffers)
```

and that seems to have worked. The buffer list is down to 34. Intero still seems to work, though I can't rule out that I've broken it subtly.

If necessary I can call that function once a day, but that's not ideal.

[...]

PR #545, opened 2018-04-16
More robustly delete network process buffers.

ChickenProp:
See #544. Unfortunately I seem to have the worst of both worlds. With
2c2546b I get "selecting deleted buffer" errors, sometimes causing emacs
to hang briefly; and with 37af015 my buffer list continues to grow,
though maybe slower than before. (I'm using emacs 25.1.1.)

It appears that for me, the "deleted\n" event is only getting triggered
inconsistently when `(delete-process)` is called. So I return to killing
the buffer immediately. But I also handle the deleted event to avoid
errors.

I considered having

    (when (buffer-live-p (process-buffer process))
      (kill-buffer (process-buffer process)))

inside the deleted event, as a fallback. But it doesn't seem to get
called for me in normal workflow, so if something unusual is happening,
I decided to not destroy the evidence of it.

### DIFF
--- a/elisp/intero.el
+++ b/elisp/intero.el
@@ -2071,15 +2071,19 @@ as (CALLBACK STATE REPLY)."
 
 (defun intero-network-call-sentinel (process event)
   (pcase event
-    ("deleted\n"
-     (kill-buffer (process-buffer process)))
+    ;; This event sometimes gets sent when (delete-process) is called, but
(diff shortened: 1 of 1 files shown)


purcell:
Thanks - this is a reasonable alternative approach, so I'll go ahead and merge it. Thanks for tracking the issue down. I suspect different behaviour of network processes between Emacs versions.

Issue #553, opened 2018-05-16
targets are not restored when intero starts upon loading a file


abailly:
I have the following project structure:
```
stack.yaml
  +-- foo
            +--- library
                       +---- Foo.hs
            +--- tests
                       +---- FooTest.hs
  +-- bar
            +--- library
                       +---- Bar.hs
            +--- tests
                       +---- BarTest.hs
  +-- .dir-locals.el
```

The `.dir-locals.el` file contains

```
;;; Directory Local Variables
;;; For more information see (info "(emacs) Directory Variables")

((haskell-mode
  (intero-targets "foo:lib" "foo:test:tasty")))
```

My `.emacs` file contains

```
(custom-set-variables
 ;; custom-set-variables was added by Custom.
 ;; If you edit it by hand, you could mess it up, so be careful.
 ;; Your init file should contain only one such instance.
 ;; If there is more than one, they won't work right.
 '(safe-local-variable-values
   (quote
    ((intero-targets "foo:lib" "foo:test:tasty" )))))
```

Restarting Emacs, opening `FooTest.hs`  and then typing `C-c C-l` to load it I observe that the test dependencies are actually not loaded and that intero starts with the following command-line:
```
  stack ghci --with-ghc ghci "--docker-run-args=--interactive=true --tty=false" --no-build --no-load --ghci-options  foo
```

instead of the expected

```
  stack ghci --with-ghc ghci "--docker-run-args=--interactive=true --tty=false" --no-build --no-load --ghci-options  foo:lib foo:test:tasty
```

I then need to manually redefine the targets (which intero asks me to save and mark as safe) for it to work.

purcell:
Indeed. It's quite hit-and-miss, and I need to spend some time investigating exactly why this happens. Similarly, changes made with `M-x intero-targets` do not always take effect reliably across the shared scope. It's somehow due to:

* Setting `intero-targets` dir-locally only for `haskell-mode`, while the intero worker buffers are not in that mode.
* Needing to propagate `intero-targets` into the buffers for intero processes and related haskell files when it changes.

but the interplay is clearly a bit fiddly.

Other open issues mention this, but let's keep this one open to track the specific problem.

abailly:
Is there a workaround I could apply?

PR #562, opened 2018-06-15
Fix restoring intero-targets from .dir-locals.el

dgaw:
Fix for https://github.com/commercialhaskell/intero/issues/553


purcell:
Oh wow, I looked for the cause of this, and didn't find that specific issue - thanks!

PR #567, opened 2018-07-13
Add option to show warnings even when errors exist

janpath:
One important use case is to be able to put a typed hole somewhere and still get warnings for the rest of the code.

### DIFF
--- a/elisp/intero.el
+++ b/elisp/intero.el
@@ -157,6 +157,11 @@ by default."
   :group 'intero
   :type '(repeat string))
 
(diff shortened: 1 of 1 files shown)


chrisdone:
What's the test case for this?

janpath:
```
{-# OPTIONS -Wmissing-signatures #-}
test = () _
```

For this file I want to get the type error arising from applying `()` to an argument, the typed-holes error and the missing-signatures warning.

chrisdone:
screenshot from 2018-07-13 15-23-44

Seems to already work fine. What version of GHC are you on?

PR #570, opened 2018-07-31
Restart GHC in intero-repl-buffer if repl buffer already exists

sergv:
This can occur if GHC died with a segfault, for example.


purcell:
Seems reasonable, thanks!

Issue #572, opened 2018-08-02
wrong-type-argument stringp #<process intero>


manuel-uberti:
Hi,

I am running latest `intero` from MELPA. After opening a `.hs` file, `intero` boots up correctly and I can switch to the REPL with <kbd>C-c C-z</kbd>.

If I try to switch back to REPL with the same key binding a second time I get this error:
```
Debugger entered--Lisp error: (wrong-type-argument stringp #<process intero>)
  intero-repl-buffer(nil t)
  intero-repl(nil)
  funcall-interactively(intero-repl nil)
  call-interactively(intero-repl nil nil)
  command-execute(intero-repl)
```
Other tries at switching back to the REPL don't work and I end up with the same error. I get the same error when trying to load the file in the REPL with <kbd>C-c C-l</kbd>.

Some details about my system, in case you should find them helpful:
```
GNU Emacs 27.0.50 (build 1, x86_64-debian-linux-gnu, GTK+ Version 3.22.30)
 of 2018-08-02

Repository revision: e65ec81fc3e556719fae8d8b4b42f571c7e9f4fc
Configured using:
--host=x86_64-debian-linux-gnu --with-modules

Emacs uptime: 2 minutes, 56 seconds
Colour theme: sanityinc-tomorrow-night
Operating system: Ubuntu 18.04.1 LTS
Window system: x11
Desktop environment: GNOME Shell 3.28.2
```
If you need further information please ask.

alexpeits:
I have the same issue, reverting #570 fixed it for me.

manuel-uberti:
Thanks @alexpeits, I'll try to do the same for now.

alexpeits:
I have a PR to fix it while keeping the functionality of #570 : #573

loupgaroublond:
I just got bit by the same issue. Any chance #573 will get merged soon?

manuel-uberti:
@loupgaroublond you could clone the repository and apply the fix in the PR while waiting for it to get merged. That's what I did for now. :)

loupgaroublond:
Of course I can, which is effectively what i've done, but I'd like to know when I can go back to tracking the master branch :).

For anyone who ended up here via Google and is wondering how to do this, I added the following to `dotspacemacs-additional-packages` in my `.spacemacs` file:

```lisp
(intero :location (recipe :repo "alexpeits/intero"
                               :fetcher github
                               :files ("elisp/intero.el")))
```

manuel-uberti:
[...]

PR #573, opened 2018-08-02
Return the buffer instead of the process when creating the repl buffer

alexpeits:
This fixes #570 . `and` blocks in elisp return the last "truthy" value, otherwise nil. #570 as a result returned the process and not the buffer when the `and` expression was true, resulting to an error described in #572 .


purcell:
Thanks

PR #581, opened 2018-10-09
Networked intero

chrisdone:
Pinging @purcell.

This is a rather substantial change to intero. I've yet to run the GHC test suites, this serves as a preview PR that I'm testing on work projects. It's not a big change to the elisp, but the executable has changed a fair bit.

# The problem

Currently, Emacs launches intero the binary as a piped process and controls it end to end, with `stack` wrapping it up with all the right packages and target choices. This provides a great user experience, but sometimes you need to launch intero outside of your regular environment. Think:

* Working on a client nix project
* Working on something inside a docker container
* Working on something inside a VM (e.g. VirtualBox)

In each of these cases, the way to invoke intero can be complicated.

# The solution

Change the way Emacs talks to intero. Make it talk over a network for everything.

* Emacs already abstracts over whether you're talking to a piped process or a network socket with the same API. This code basically didn't have to change. I simply changed the launcher to either:
  1. If Intero is already running on a port, just connect to that instead with `make-network-process`.
  2. If intero isn't running, then start it ourselves with the existing `start-file-process` call.

I've tested this with an intero invocation inside a docker-compose project, with port 1234 exposed. I specify `INTERO_SERVICE_PORT=1234` and intero starts, listens on that port, and writes `1234` to `.intero/port`. Emacs automatically checks for this file and auto-connects to that process. 

You have to make sure that Emacs and the docker/vm container share the same directory roots with e.g. `docker run -v$(pwd):$(pwd)`.

As for the intero binary, the main change is that all commands will either write to stdout or a handle to a socket. Additionally, compiler messages are captured and written to the socket instead of directly to stdout/stderr.

This also represents a step forward to moving away from pretending to be a commandline app and closer to being a networked service. It also represents a step toward something like Microsoft's standard IDE protocol. But I'd hope by that point we can all switch to HIE.

# Remaining work

 Some stand-out issues:

* The Emacs lisp code could use the concept of "workers" already built into intero's internal API, and do away with the "network call" that I implemented earlier this year (or last?). Instead, we'd have the regular `'backend` worker and then a `'readonly` worker which does readonly commands (querying).
* We may have to update tramp support, or consider removing it and recommending rsync/sshfs.
* Testing on windows?
* Test on various GHCs.
* If we go forward with this approach, we ought to make it continue to work in the usual piped fashion, so that upstream users (vim, VS Code, etc.) don't have to make any changes.

### DIFF
--- a/elisp/intero.el
+++ b/elisp/intero.el
@@ -157,6 +157,11 @@ by default."
   :group 'intero
   :type '(repeat string))
 
+(defcustom intero-port-file
+  ".intero/port"
+  "File that stores the port number of a running service."
+  :group 'intero)
+
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 ;; Modes
 
@@ -322,9 +327,6 @@ This is slower, but will build required dependencies.")
 
 (defvar-local intero-starting nil
   "When non-nil, indicates that the intero process starting up.")
-
-(defvar-local intero-service-port nil
-  "Port that the intero process is listening on for asynchronous commands.")
 
 (defvar-local intero-hoogle-port nil
   "Port that hoogle server is listening on.")
@@ -1816,7 +1818,7 @@ x:\\foo\\bar (i.e., Windows)."
 (defun intero-get-type-at-helper-process (beg end)
   (replace-regexp-in-string
    "\n$" ""
-   (intero-blocking-call
+   (intero-blocking-network-call
     'backend
     (intero-format-get-type-at beg end))))
 
@@ -1916,7 +1918,7 @@ type as arguments."
   "Make the blocking call to the process."
   (replace-regexp-in-string
    "\n$" ""
-   (intero-blocking-call
+   (intero-blocking-network-call
     'backend
     (format ":loc-at %s %d %d %d %d %S"
             (intero-path-for-ghci (intero-temp-file-name))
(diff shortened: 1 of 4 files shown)


purcell:
Cool. My concern is the extra complication of building network machinery in - given a way to customise the command that executes intero, there would already be ways to have the command run it remotely, or in a docker container etc. BTW, a nice network-centry model for language interaction systems is `cider`, for Clojure. Might be worth taking a look at how they do things.

One other thing I'm thinking is: if you're going to have `intero` talk on a network, have you considered making it speak the language server protocol? I increasingly feel that that's the way forward, and to have something like the haskell language server that can self-install would be pretty nice. /shrug

chrisdone:
> My concern is the extra complication of building network machinery in 

I think I'm in agreement on not introducing extra complication. I've realised I don't feel like maintaining something like this further, so I'm abandoning this branch.

> given a way to customise the command that executes intero, there would already be ways to have the command run it remotely, or in a docker container etc. 

It's true, I tried that approach to limited success. I think a generalized program that can act as a network proxy between two piped ends would be handy for that, to avoid the overhead of setting up a full-blown SSH server.

> BTW, a nice network-centry model for language interaction systems is cider, for Clojure. Might be worth taking a look at how they do things. One other thing I'm thinking is: if you're going to have intero talk on a network, have you considered making it speak the language server protocol? I increasingly feel that that's the way forward, and to have something like the haskell language server that can self-install would be pretty nice. /shrug

I've considered that it would be good, I looked over the protocol spec. I have zero interest in doing the work.

purcell:
> I think a generalized program that can act as a network proxy between two piped ends would be handy for that, to avoid the overhead of setting up a full-blown SSH server.

socat?

> I have zero interest in doing the work.

Haha, fair enough. :-)

PR #589, opened 2018-11-02
Travis: cache .stack dir

develop7:
cache `$HOME/.stack` dir (hopefully) improving consecutive build times

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -47,3 +47,6 @@ before_install:
 
 script:
 - stack $ARGS --no-terminal test --ghc-options=-Werror
+
+cache:
+- '~/.stack'
--- a/.travis.yml
+++ b/.travis.yml
@@ -49,4 +49,5 @@ script:
 - stack $ARGS --no-terminal test --ghc-options=-Werror
 
 cache:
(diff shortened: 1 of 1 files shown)


develop7:
@chrisdone pls trigger rebuild in Travis so we could see if cache does actually bring any improvement

purcell:
I restarted the [previous build](https://travis-ci.org/commercialhaskell/intero/builds/449851012?utm_source=github_status&utm_medium=notification), which had taken 36m.

develop7:
OK, so results:

| Build | Time |
| --- | --- |
| [cached build](https://travis-ci.org/commercialhaskell/intero/builds/449851012) | 12:24 |
| [master branch](https://travis-ci.org/commercialhaskell/intero/builds/449842512) | 28:04 |

develop7:
Please note the caches are per-branch (along other "per-"s) so no immediate speedup after merging.

purcell:
I'll merge this, but note that the Travis caches can get pretty big, and at that point they often become so slow to store and fetch that they make the build slower, not faster. So let's keep an eye on how this works out in practice.

PR #591, opened 2018-11-06
Travis: make use of build matrix

develop7:
Our case with all possible combinations of resolver and OS and only one exclusion is quite an exemplary case for the build matrix, so this converts explicit jobs to implicit.

Also converts `ARGS` environment variable to `RESOLVER` since we're using only it anyway.


chrisdone:
:+1: Great!

PR #595, opened 2018-12-21
travis: add nightly Stackage, reorder resolvers

develop7:
Builds against nightly do not trigger build failure though.


chrisdone:
Thanks!

PR #598, opened 2018-12-23
GHC 8.6.3

DanBurton:
I do not claim to understand exactly what's going on here, but I do claim that this version of the test suite passes with all of the following builds:

```
$ stack test
$ stack --resolver lts-12.25 test
$ stack --resolver lts-13.0 test
```

I also did some very rudimentary manual testing via

```
$ stack --resolver lts-13.0 build --copy-compiler-tool
```

and it seems to be working as intended.

### DIFF
--- a/intero.cabal
+++ b/intero.cabal
@@ -73,7 +73,7 @@ executable intero
     bytestring,
     directory,
     filepath,
-    ghc >= 7.8 && <= 8.6.1,
+    ghc >= 7.8 && <= 8.6.3,
     ghc-paths,
     haskeline,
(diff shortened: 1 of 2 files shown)


chrisdone:
Cheers! Will push to hackage.

chrisdone:
Pushed to hackage. When it's on stackage people can update their emacs lisp to master which references intero 0.1.35.

chrisdone:
Or `M-x customize-variable`  of `intero-package-version` if you had customized it. To 0.1.35.

PR #601, opened 2019-01-07
Substitute command keys for flycheck message

yuhan0:
The flycheck message for a suggestion is hardcoded to
"Hit `C-c C-r` in the Haskell buffer to apply suggestions"

But this may be misleading depending on the actual keybindings - e.g. in Spacemacs `C-c C-r` is bound to haskell-process-reload instead.


purcell:
Nice, thanks.

PR #605, opened 2019-01-21
Add libtinfo installation instructions for Fedora derivatives

manenko:
### Reason

The `README.md` uses `libtinfo-dev` as a package name of `libtinfo` for GNU/Linux. While the name is valid for Debian-based distros, other distros put the library into a different package. 

### Changes

- Explicitly mention `libtinfo-dev` as a package for Debian derivatives.
- Add package name that installs `libtinfo` on Fedora derivatives.

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,9 +25,12 @@ Key binding | Description
 
 ## Install requirements
 
-Make sure you have this package installed on Linux:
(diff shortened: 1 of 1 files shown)


chrisdone:
Cheers!

PR #609, opened 2019-03-08
Bump GHC upper bound to 8.6.4

korayal:
Well, nothing much to say. The upgrade from 8.6.3 to 8.6.4 is pretty much about the change in the upperbound.


chrisdone:
Cool, will merge once it passes CI.

PR #619, opened 2019-05-11
Bump GHC upper bound to 8.6.5

develop7:
Following up the GHC upgrade in stackage LTS-13.20


chrisdone:
Looks like the tests pass. Great!

PR #623, opened 2019-05-23
Bump intero package version in intero.el

### LINKED ISSUE #622: Outdated Intero 0.1.39 for GHC 8.6.5
## Problem

When installed from MELPA, the default version for intero's autoinstall is still set to `0.1.39` which doesn't support GHC `8.6.5`. Since Intero `0.1.40` itself already supports GHC `8.6.5`, I simply upgraded the version to autoinstall in `intero.el` to fix. (related to #618 for supporting `GHC 8.6.5`)

### Input program/project

N/A

### Expected behavior

Successful autoinstall of `Intero 0.1.40` for `GHC 8.6.5`.

### Actual behavior

After install from MELPA, intero's init installs version `0.1.39` which results in a version mismatch with GHC `8.6.5` (upper bound at `8.6.4` I believe)

### Workarounds

Change version to `0.1.40` as set in `intero-package-version` in `intero.el`.

schuermannator:
Fix #622

### DIFF
--- a/elisp/intero.el
+++ b/elisp/intero.el
@@ -72,9 +72,9 @@
 (defcustom intero-package-version
   (cl-case system-type
     ;; Until <https://github.com/haskell/network/issues/313> is fixed:
-    (windows-nt "0.1.39")
-    (cygwin "0.1.39")
-    (t "0.1.39"))
+    (windows-nt "0.1.40")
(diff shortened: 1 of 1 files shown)


chrisdone:
:+1:

PR #628, opened 2019-06-16
Use `fno-code fwrite-interface` when GHC >= 8.4.1

### LINKED ISSUE #627: Consider -fno-code -fwrite-interface in emacs
I gather from [here](https://www.reddit.com/r/haskell/comments/8kn1x7/ghcid_for_the_win/dzat57z/), that a major reason that intero processes in emacs use `-fobject-code` is because it reloads faster than a pure `-fno-code`. I recently learnt about `-fwrite-interface` from @mpickering. This makes `-fno-code` reloading fast, and it beats `-fobject-code` as one would expect.

There may well be other reasons to use `fobject-code` over `fno-code` that I'm not aware of, but I thought I would mention.

If you'd rather not have this as a default of intero's emacs mode, perhaps it could be configurable?

rjmk:
This should fix #627.

A couple of points/questions:

* Does this mean that other editor integrations now need to follow the switching logic, on pain of intero not correctly tracking when files have been loaded?
* I've checked that the elisp compiles, but I'm afraid I'm too much of an emacs noob to test this locally. I tried `M-x install-file` but got errors when intero tried to install itself. Sorry about that.

> Before this commit, we pass `-fobject-code` to ghci, which produces
> build artifacts (object files). This allows subsequent reloads to be
> faster and improves the performance of intero.
> 
> It is possible to output interface files while only doing the work of
> typechecking with the flags `fno-code` and `fwrite-interface`. This,
> again, improves the performance of intero.
> 
> Unfortunately, `fno-code` does not work well with Template Haskell
> before GHC version 8.4.1. Therefore, we switch which flags we provide
> based on version number.

### DIFF
--- a/elisp/intero.el
+++ b/elisp/intero.el
@@ -1883,9 +1883,12 @@ type as arguments."
                  (intero-async-call
                   'backend
                   (concat ":load " (intero-path-for-ghci (intero-temp-file-name)))))
-               (intero-async-call
-                'backend
-                ":set -fobject-code")
+               (mapc
+                 (lambda (flag)
+                   (intero-async-call
+                    'backend
+                    (append ":set " flag)))
+                 (intero-ghci-output-flags))
                (replace-regexp-in-string
                 "\n$" ""
                 (intero-blocking-call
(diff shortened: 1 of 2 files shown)


elisp/intero.el
```diff
@@ -2660,6 +2666,14 @@ For debugging purposes, try running the following in your terminal:
         (setq intero-ghc-version
               (intero-ghc-version-raw)))))
 
+(defun intero-ghci-output-flags ()
+  "Get the appropriate ghci output flags for the current GHC version"
+  (with-current-buffer (intero-buffer 'backend)
```
rjmk: I did just cargo cult this `with-current-buffer`. Seems like it might be needed for `intero-ghc-version`

elisp/intero.el
```diff
@@ -2660,6 +2666,14 @@ For debugging purposes, try running the following in your terminal:
...
+    (let ((current-version (mapcar #'string-to-number (split-string intero-ghc-version "\\."))))
```
rjmk: Do I get away with the direct reference to the variable here? I could call out to the function instead

chrisdone: That looks fine.

chrisdone:
I'll test this out. Thanks! I was on vacation.

PR #641, opened 2019-11-03
Add Ghcide as a possible alternative

ndmitchell:
Thanks for your awesome work on Intero. Some of Ghcide was bootstrapped by taking Intero code, so we greatly appreciate it. Perhaps list Ghcide as a possible alternative? (In future times it may merge with HIE, but that's probably a way off, and not a guarantee)


chrisdone:
Ta! :+1:

PR #642, opened 2019-12-22
replace append by concat for flag generation

mikaeldusenne:
Hello,
This is my first participation in an open source project, so please tell me if I'm doing anything wrong :)

I tracked this bug from this (see my answer):
https://stackoverflow.com/questions/59058854/intero-error-wrong-type-argument-stringp-nil

NB I didn't quite solve the original problem (intero-ghc-version not being set immediately), but this one was easier to pinpoint.

The bug was introduced by 0a87ecaf98950eae7516613ffc84f901bbf48f54 

I hope this helps!

Best,
Mikael

### DIFF
--- a/elisp/intero.el
+++ b/elisp/intero.el
@@ -1886,7 +1886,7 @@ type as arguments."
                  (lambda (flag)
                    (intero-async-call
                     'backend
-                    (append ":set " flag)))
+                    (concat ":set " flag)))
                  (intero-ghci-output-flags))
                (replace-regexp-in-string
                 "\n$" ""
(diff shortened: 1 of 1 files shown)


mikaeldusenne:
Hi again,
I think I found out the reason why `intero-ghc-version` was nil,
see my second commit, and also the consolidated answer I posted in the [SO thread ](https://stackoverflow.com/questions/59058854/intero-error-wrong-type-argument-stringp-nil)

```
(split-string intero-ghc-version "\\.")
```

should really be

```
(split-string (intero-ghc-version) "\\.")
```

so that the local variable is set properly, right ?

With these two fixes I am back with a running intero :)

tr00per:
Seems that #643 contains all the commits from this Pull Request, so this can be closed. Also - @chrisdone would have to do something with the failing CI on master first, but it was 2 months since the last build :/

PR #643, opened 2019-12-23
Hotfix/flags intero version check

mikaeldusenne:
Related to #642 , it seems the function comparing the ghc version has been used in the wrong direction.

### DIFF
--- a/elisp/intero.el
+++ b/elisp/intero.el
@@ -1886,7 +1886,7 @@ type as arguments."
                  (lambda (flag)
                    (intero-async-call
                     'backend
-                    (append ":set " flag)))
+                    (concat ":set " flag)))
                  (intero-ghci-output-flags))
                (replace-regexp-in-string
                 "\n$" ""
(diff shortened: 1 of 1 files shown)


wraithm:
I can verify that this patch works for me.

It looks like intero isn't going to be maintained here. This is really unfortunate that intero is going away... Honestly, this has been the best tool that I've used for editing Haskell ever.

@chrisdone Would you be interested in giving some of us access to the "commercialhaskell/intero" (which is what melpa references) repo, such that we can keep maintaining it there?

afcady:
So this worked for me too.

For melpa users: to get the patch running, I just did this:

```
git clone  https://github.com/chrisdone/intero
cd intero
git pull origin refs/pull/643/head
cp elisp/*.el $HOME/.emacs.d/elpa/intero-20191103.1239/
rm $HOME/.emacs.d/elpa/intero-20191103.1239/*.elc
```

and then `M-x restart-emacs` fixed it.

Also, the error message that indicates the problem is `wrong-type-argument stringp nil`.  The error message is reproduced again on every `M-x intero-restart`

